PR #25, opened 2023-07-31
Notifications adjustments. DeployActions. Docs updates

andrey18106:
- [x] DeployActions interface
- [x] Notifications adjustments (receive dynamic parameters to build known subjects)
- [x] Write short docs about Notifications


bigcat88 (approved):

bigcat88:
Why we need both `subject` and `rich_subject` ,   `message` and `rich_message` ?
What has higher priority? Can we simplify it ti be only `rich_subject` and `rich_message`, as I understand we can pass there usual strings and not only `rich` ones.

@andrey18106

bigcat88:
Merging this as nc_py_api PR needs this now

PR #29, opened 2023-08-02
ExApp version check

andrey18106:
- [x] Resolve conflicts
- [ ] Merge into auth throttling PR first

### DIFF
--- a/.github/workflows/tests-deploy.yml
+++ b/.github/workflows/tests-deploy.yml
@@ -48,6 +48,13 @@ jobs:
           repository: nextcloud/server
           ref: ${{ matrix.server-version }}
 
+      - name: Checkout Notifications
+        uses: actions/checkout@v3
+        with:
+          repository: nextcloud/notifications
+          ref: ${{ matrix.server-version }}
+          path: apps/notifications
+
       - name: Checkout AppEcosystemV2
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
         with:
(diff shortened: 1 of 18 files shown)


bigcat88:
Thoughts:
1. Admin do not stops `ExApp` and updates it manually(with new Docker image).
2. New version of `ExApp` wants to register new stuff, it did not do that in this case.
3. App already got displayed as updated in NC, it can not be updated automaticly, as we just `imitate update here`.

bigcat88:
First solution:

1. Got another version then expected
2. We disable app
3. We enable app, accepting all API scopes automatically (as with flag `--force-scopes`)

Pros:
* easy to manual update application

Cons:
* app can imitate that it's version changed, and rerequest some API scopes that was not granted to it.

bigcat88:
Second solution:

1. Got another version then expected
2. We disable app immidiatly and notify admin about disabling(how can we know who shuold be notified?)

Pros:
* admin in this case, just should enable app in the UI or console and that's all

Cons:
* We do not know know whom can be notified about this. It is ugly to create such field only for the AppEcosystem. If we imagine such field will present in the Nextcloud ❤️


@julien-nc

julien-nc:
> how can we know who shuold be notified?

All members of the admin group should be fine.

It feels weird to automatically accept new scopes from an ExApp. It makes the scope system useless when an ExApp is updated manually outside of NC.
I would go for the second solution.

Also, we could recommend not to manually update ExApps as it can lead to longer app downtimes (until an admin validates the new scopes).

Also, it is very likely that the person who updates an ExApp manually is a NC admin and knows it's wise to check the "app settings v2" to validate new scopes. So even if the notification is not perfect and can be seen late, it's good enough for a use case that is not recommended. We can decide people updating ExApps manually should know what they are doing and it's fine if the workflow is not perfect.

bigcat88 (changes requested):
Till this got rewritten to new algo

bigcat88:
will fix tomorrow, need to extend nc_py_api by adding two new APIs for extra applications.

PR #49, opened 2023-08-21
docker-aio-actions draft, commands updates

andrey18106:
Merge after changes and API sync with AIO


bigcat88 (approved):

andrey18106:
Not needed as AIO will provide docker-socket-proxy, we'll have to use just auto DaemonConfig setup if in AIO and a few general adjustments (will be done in another PR).

PR #52, opened 2023-08-22
fixed test for last nc_py_api version

bigcat88:
Last nc_py_api use `nc.notifications` instead of `nc.users.notifications`


bigcat88:
please merge after review

PR #60, opened 2023-08-29
STT & Text processing APIs

andrey18106:
Attempt to implement dynamic registration of Speech-To-Text and Text processing APIs.

Due to technical obstacles one unresolved task left as there is no possibility in PHP to override `::class` string, which is used for retrieving classes, implementing API interfaces widely used in Nextcloud.

- [x] Find out a way to properly resolve anonymous classes
- [x] Extract migration to a new one (as the previous one already released)
- [x] Add check for feature availability (since 29.0.0)

We can proceed with these APIs implementation after the following server PRs accepted and finished:

- [x] https://github.com/nextcloud/server/pull/41087
- [x] https://github.com/nextcloud/server/pull/41088
- [ ] Machine translation PR to be created


andrey18106:
Possible solution found with Julius, will update this PR later

andrey18106:
Current Text processing APIs in core requires adjustment in provider ID to use `getId()` method to identify unique provider registration like in other APIs, e.g. DashboardWidget (`IWidget`). Therefore current implementation would work only with single registration as in core it relies on `$provider::class` which in our case is anonymous class pointing to the line of code where it was created - so that multiple instances will not be handled properly.

bigcat88 (changes requested):
we will implement all this in each separate PRs

andrey18106:
STT and Text-Processing are done and merged in separate PRs. Machine translation needs server side changes the same as for STT and Text-Processing, waiting for them and after that will add this API support here.

PR #62, opened 2023-08-30
to_gif without framework [example]

bigcat88:
initial commit, all except WebDav is done


bigcat88:
Example is without any error handling or error checks, as it will require to many additional code, and should not be done in Apps by hands without special needs.

andrey18106 (approved):

PR #63, opened 2023-08-30
small general adjustments for AIO compatibility

andrey18106:


bigcat88:
@andrey18106 Is this Pull Request marinating in a state of pre-critique suspense, or it is ready to be in the spotlight of review?

andrey18106:
> @andrey18106 Is this Pull Request marinating in a state of pre-critique suspense, or it is ready to be in the spotlight of review?

Added updated AIO and Docker stuff according to latest changes, now it can be reviewed.

lib/Migration/DataInitializationStep.php
```diff
@@ -26,5 +32,13 @@ public function run(IOutput $output): void {
 		} else {
 			$output->warning('Failed to initialize data for App Ecosystem V2');
```
bigcat88, line 33: maybe here make it an `error` level, as it makes AppEcosystem unusable.

bigcat88 (approved):
Overall looks like solid one.

lib/Migration/DataInitializationStep.php
andrey18106: There is no error message type in `IOutput` [interface](https://github.com/nextcloud/server/blob/master/lib/public/Migration/IOutput.php), it's used to write messages in CLI [during installation](https://github.com/nextcloud/server/blob/aba3f99e5efa02c33fec00c55b207769762cc976/lib/private/Installer.php#L165).

lib/DeployActions/AIODockerActions.php
```diff
@@ -0,0 +1,89 @@
...
+use OCA\AppEcosystemV2\Service\DaemonConfigService;
+use OCP\IConfig;
+
+/**
+ * Class with utils methods for AIO setup
+ */
+class AIODockerActions {
+	public const AIO_DAEMON_CONFIG_NAME = 'docker_aio';
+	public const AIO_DOCKER_SOCKET_PROXY_HOST = 'nextcloud-aio-docker-socket-proxy:2375';
+	public const AIO_DAEMON_HOST = 'http://nextcloud-aio-apache:11000/index.php';
```
szaimen, line 17: Just a remark herr: it would probably better to point this at https://$NC_DOMAIN as port 11000 is not always choosen but the above should usually work always...

szaimen: Though probably not on your dev instance since you are running with self-signed cert...

szaimen: For production instances AIO always requires a valid cert for the chosen domain and always runs on the default port 443. that is why the above should work.

andrey18106: Hmm good points. @bigcat88 What do you think?

andrey18106: > For production instances AIO always requires a valid cert for the chosen domain and always runs on the default port 443. that is why the above should work.

So locally for testing we could always create our own daemon and user apache proxy for ExApps to communicate to nextcloud, instead of this automatically created. Thanks for pointing on this)

PR #69, opened 2023-09-11
authorization-app-api, auth simplification

andrey18106:


lib/AppAPIAuthBackend.php
```diff
@@ -28,7 +28,7 @@ public function __construct(
 	public function check(RequestInterface $request, ResponseInterface $response) {
 		if ($this->request->getHeader('AE-SIGNATURE')) {
```
bigcat88, line 29: What is this? Should `AUTHORIZATION-APP-API` be here instead?

bigcat88 (approved):

PR #77, opened 2023-09-15
UI implementation(appstore, daemons)

andrey18106:
- [ ] polish
- [ ] update docs with screenshots


lib/DeployActions/DockerActions.php
```diff
@@ -556,6 +574,14 @@ private function buildDefaultExAppVolume(string $appId): array {
 		];
 	}
 
+	public function isDockerSocketAvailable(): bool {
+		$dockerSocket = '/var/run/docker.sock';
+		if (file_exists($dockerSocket) && is_readable($dockerSocket)) {
```
bigcat88, line 579: maybe here should be `is_writable `?

bigcat88:
`registerDefaultDaemonConfig` should perform check if there is docker socket available.

all other we discussed in a call and is fine for first Beta Public version.

bigcat88 (approved):

PR #78, opened 2023-09-29
fixed an url for CalDav, CardDav and other extension on top of WebDAV

bigcat88:
subj, currently they are not working with AppAPI auth.


bigcat88:
deploy errors unrelated

PR #96, opened 2023-10-18
ExApp init, disable apps actions if daemon not accessible

andrey18106:


bigcat88 (approved):
Add "sleep" in cycle and let's merge, all bug fixes if found will be in next PR if needed.

PR #97, opened 2023-10-20
use postAsync or event to enable ExApp after its initialization

andrey18106:
Using asynchronous event after ExApp finished initialization to enable it.

- [ ] Figure out the way to make it work with single process webserver (like in CI tests built-in PHP webserver), because of asynchronous operations used.


bigcat88 (changes requested):
After checking with additional logs, it was revealed that both proposed solutions do not help the situation.

The version with the event is definitely going in sync.

Why the option with `postAsync` works synchronously (even if we remove “then” and “wait”) is not yet clear.

For now, we’ll correct the situation by simply increasing the number of PHP processes in test actions from one to two.

PR #109, opened 2023-11-07
fixed UI error when default_daemon_config config missing

bigcat88:
When a daemon has been selected by default and then for some reason its configuration is missing from the DB, the AppAPI application becomes unusable from the UI.

_I came across this locally today._


andrey18106:
Should we also adjust that here: https://github.com/cloud-py-api/app_api/blob/main/lib/Controller/ExAppsPageController.php#L107?

bigcat88:
> Should we also adjust that here: https://github.com/cloud-py-api/app_api/blob/main/lib/Controller/ExAppsPageController.php#L107?

yeap

andrey18106 (approved):

PR #110, opened 2023-11-07
ExApp embedded iframe draft

andrey18106:


bigcat88 (changes requested):
we should choose alternative way for this task, to allow **nextcloud/vue** to be easy used in ExApps

PR #112, opened 2023-11-08
add internal methods for requests to ExApps by appid

andrey18106:


bigcat88:
now it looks really solid.

PR #126, opened 2023-11-25
manual_install: added Note that deploy/start should be handled by dev

bigcat88:
Subj. 

_I thought that this was already clear, but according to reviews, not everyone understands this, so it’s better to explicitly indicate it in the documentation_

@kyteinsky or should we also add this note in some other places?


bigcat88:
image

kyteinsky (approved):
looks perfect!

andrey18106 (approved):

Issue #117, opened 2023-11-10
GPU passthrough not working (NVIDIA)


bigcat88:
We need this for complex models.

Theoretically this should work and we took code from `AIO` for this(for NVIDIA at least), but during tests we found that do not work in our current setups.

And after we solve this, all this stuff should be documented.

_Andrey will check on **Nvidia**, he will install a separate Linux on his system for this, I will check on **AMD**.
We will check everything not on a virtual machine, but on separate Linux installed for this purpose on real equipment, as it should be in production._

PR #130, opened 2023-11-27
Adjustments of NVIDIA GPU support in Docker containers

andrey18106:
Resolves: #117


bigcat88 (approved):

bigcat88:
Is this still will work with current https://github.com/nextcloud/all-in-one/blob/main/Containers/docker-socket-proxy/haproxy.cfg ?

andrey18106:
> Is this still will work with current https://github.com/nextcloud/all-in-one/blob/main/Containers/docker-socket-proxy/haproxy.cfg ?

Not tested yet. It might require additional investigation and adjustment on AIO side (e.g. installing required libs for nvidia containers support in Docker).

Issue #169, opened 2023-12-19
Nextcloud_url invalid default value, missing postparsing


bigcat88:
NEXTCLOUD_URL="http://nc.com/index.php/"

is invalid for ExApps.

Supported urls should be only:

http://nc.com/index.php
http://nc.com

Slash at the end should not be supported by ExApps, so it is an AppAPI task to control this.

PR #170, opened 2023-12-19
fix: remove last slash in url

andrey18106:
Resolves: #169


bigcat88 (approved):
I added additional check on backend and adjusted tests to test backend part

re review yourself, maybe I missed something

Issue #171, opened 2023-12-19
Missing "Uninstall" button for ExApp without refreshing page


bigcat88:
Without refreshing page, the button is missing.

![image](https://github.com/cloud-py-api/app_api/assets/13381981/464ed910-fb1a-4cb7-9282-3babc1973ff1)

**When I refresh page, I see it.**

PR #177, opened 2023-12-20
ExApps management and Admin settings UI logic fixes

andrey18106:
Resolves: #171 

Also fixes:

* Update button in UI not working.


bigcat88:
image

`Failed to enable ExApp` message do not go away after successful enabling.

Steps to reproduce:

1. Enable App that for example is turned off(or any other way to make Enabling failed)
2. Again Enable App, but now make enabling process be successful.
3. Message from Step1 does not go away, even if Step2 finish with success.

bigcat88 (changes requested):
And with this PR for some reason I cannot update an App from UI, pressing Update button, only disables ExApp and raises an error.

bigcat88:
> And with this PR for some reason I cannot update an App from UI, pressing Update button, only disables ExApp and raises an error.

this was fixed by last commit to this PR.

_bug was introduced in previous releases that added GPU support_

bigcat88 (approved):
now it is fine

PR #198, opened 2024-01-05
fix FileAction register logic depending on icon

andrey18106:
Fixing missed bug introduced in recent filesplugin update: when no icon set - FileAction is not registered in NC28.


bigcat88 (approved):
I hope you tested it :)

PR #199, opened 2024-01-05
delay userId querying from DI later in anonymous class

andrey18106:
During PHP App registration process we don't have a complete session and other stuff queried from DI which is for some reason (I don't know why yet) breaks its usage later.


bigcat88 (approved):
let's publish a new version with this, as this bug is critical.

PR #210, opened 2024-01-13
MachineTranslation

andrey18106:


bigcat88:
should we move all AI Providers controllers in a separate folder?

lib/Controller/MachineTranslationController.php
```diff
@@ -0,0 +1,129 @@
...
+		$this->request = $request;
+	}
+
+	#[NoCSRFRequired]
+	#[PublicPage]
+	#[AppAPIAuth]
+	public function registerProvider(
+		string $name,
+		string $displayName,
+		string $fromLanguages,
```
bigcat88, line 43: probably "arrays" here?

andrey18106: Yeah, missed that, likely we should expect array to encode it later to save in DB.

bigcat88 (approved):

Issue #163, opened 2023-12-18
I18N for Notifications (and other futher message directly to one user)


StefanRichterHuber:
### Describe why it is important and where it will be useful

 Nextcloud each user can set the applications language to have the best user experience and lowest entry barrier. 
Currently the available APIs give no indication on  the preferred language for the user.

### Describe your proposed solution

Possible solutions:

1) Send an additional `X-USER-LANG` header with each request for an user to given the app the possibility to react in the users language.
Pro: Downwards compatible. Con: App has to keep track of the user language internally
2) Extend the [Get list of NC users](https://cloud-py-api.github.io/app_api/tech_details/api/utils.html) endpoint with additional metadata per user, especially language but also user group for additional authorization.
Pro: Best extensible place to put additional user metadata. Con: Not downwards compatible.
3) Enhance the [Notification API](https://cloud-py-api.github.io/app_api/tech_details/api/notifications.html) to accept `rich_subject` and `rich_message` in multiple languages and let the ui choose which one. Pro: Works without any additional information about the user. Con: Not downwards compatible, depending on the implementation.

```json
{
        "params": {
                "object": "app_api",
                "object_id": "app_api_id",
                "subject_type": "app_api_ex_app",
                "subject_params": {
                        "rich_subject":  [
                               {
                                       "lang": "en",
                                      "text": ""Image {file} successfully upscaled!"
                              }, 
                               {
                                       "lang": "de",
                                      "text": ""Bild {file} erfolgreich vergrößert!"
                              }, 
                         ],
                        "rich_subject_params": {
                                "file": {
                                        "type": "file",
                                        "id": 123,
                                        "name": "upscaled_image_name",
                                        "path": "path/to/upscaled_image_name"
                                }
                        },
                        "rich_message"://  ... see rich subject
                        "rich_message_params": {
                                "user": {
                                        "type": "user",
                                        "id": "admin",
                                        "name": "admin"
                                }
                        },
                        "link": "http(s)://nextcloud.local/index.php/apps/files/?fileid=123"
                }
        }
}

```

bigcat88:
Implemented in AppAPI `2.2.0`

Issue #162, opened 2023-12-18
I18N for File actions menu (Top Menu Entry and Talk bots)


StefanRichterHuber:
### Describe why it is important and where it will be useful

In Nextcloud each user can set the applications language to have the best user experience and lowest entry barrier. Unfortunately the current implementation of the registration of File  Actions Menu, Top Menu Entry and Talk Bots does only support to register a menu in one single language without any hint which is the (preferred) language of the Nextcloud instance.

### Describe your proposed solution

Enhance the registration of menu entries like that. If you check if displayName is a plain string or an object this solution is downwards compatible,too.

```json
{
        "name": "unique_name_of_file_actions_menu",
        "displayName": [
            {
                "lang": "en",
                "text": "Display name (for UI listing)"
            },
            {
                "lang": "de",
                "text": "Anzeige name"
            },
            {
                "lang": "default",
                "text": "Display name (for UI listing)"
            }
        ],
        "actionHandler": "/action_handler_route"
        "mime": "mime of files where to display action menu",
        "icon": "img/icon.svg",
        "permissions": "permissions",
        "order": "order_in_file_actions_menu",
}

```

bigcat88:
_Questions related to translation support for ExApp_

We know about this, perhaps this will be our next task, but specifically in this case it is not us who decide.

At the moment, we have not yet decided how ExApp will support translations...

PR #227, opened 2024-02-09
feat(l10n): L10N support for ExApps

andrey18106:
This PR introduces support of ExApps translations and includes only changes in AppAPI.
The core server part is kept the same.

ref1: #162 
ref2: #163


lib/Controller/ExAppsPageController.php
```diff
@@ -451,6 +451,7 @@ public function enableApps(array $appIds, array $groups = []): JSONResponse {
 						}
 					} else {
 						$this->logger->error(sprintf('Failed to deploy %s ExApp', $appId));
+						$this->exAppService->removeExAppL10NFolder($appId);
```
bigcat88, line 454: we do not need this, extraction of `L10N` folder should be during "Registration" step at the end

bigcat88 (approved):

Issue #219, opened 2024-01-28
 Failed to deploy ExApp 


ccgauvin94:
### Describe the bug

I've installed AppAPI. I've also installed the docker proxy. I've successfully registered it as the daemon. If I open a terminal in the proxy container, I can successfully touch /var/run/docker.sock. When I go to the store and try to install any of the 5 currently available apps through the GUI, I get this error:

` Failed to deploy ExApp `

That's all I see in the GUI. In the NextCloud log, the only line that seems relevant is:

`[Sun Jan 28 03:31:25.883881 2024] [php:notice] [pid 51] [client 172.20.0.4:36082] cURL error 18: transfer closed with outstanding read data remaining (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://nextcloud_docker_socket_proxy:2375/v1.41/images/create?fromImage=ghcr.io/cloud-py-api/ai_image_generator_bot:2.0.0`

There's no log output from the proxy

### Steps/Code to Reproduce

Select "Deploy and Enable" on an ExApp. 

### Expected Results

Should install and enable. 

### Actual Results

Seemingly fails to download. 

### Setup configuration

Docker compose file for the socket proxy:
  nextcloud_docker_socket_proxy:
    image: ghcr.io/cloud-py-api/aa-docker-socket-proxy:release
    container_name: nextcloud_docker_socket_proxy 
    privileged: true
    restart: always
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    environment:
      - NC_HAPROXY_PASSWORD="some_secure_password"
    networks:
      - nextcloud_network

bigcat88:
Thanks for reporting it, cause I totally missed about this(I saw such behaviour in one test production setup two days ago), but was ovehelmed with other stuff and bugs.

Will try to find out today why is the deploy process broken with Docker Socket Proxy

bigcat88:
<img width="911" alt="image" src="https://github.com/cloud-py-api/app_api/assets/13381981/e4d1977f-3e76-488f-9685-ba2564a9ccf8">

<img width="1404" alt="image" src="https://github.com/cloud-py-api/app_api/assets/13381981/3518bd73-b9ce-4de5-a716-82d7dc6b2721">

<img width="252" alt="image" src="https://github.com/cloud-py-api/app_api/assets/13381981/0816e185-98ce-407d-a73b-016f2dd034d6">


@andrey18106 Any ides what is this? 
[...]

PR #233, opened 2024-02-16
"Deploy" removed, Register Rework

bigcat88:
PR should not break anything, for old parts added a workarounds, that we will remove in a month when all ExApps will publish their updates.

**Changes**:

* UI now use the same algorithm/code for ExApp `register` & `update` as the CLI commands.
* Deprecated "deploy" command, now `register` commands performs deploy.

**Refactoring**:

* Removed internal `DispatchInit` command, as from CLI we already can do it without spawning additional process.
* Removed hack(`status['active']`) when we have some half-enabled state of ExApp when it is not enabled, but already can call APIs, now ExApp enables before calling `/init`
* Made code more consistent in many place.

OTHER CHANGES FOR DEVS:

* For `--json-info` parameter in occ `register` command keys renamed: "appid" -> "id", "system_app" -> "system"

    _Old naming is still supported but will be removed in future, in such way we make consistent parameters between `info.xml` and  `--json-info`._

----

Related: #219 
_AppAPI will perform deploy & registration in a background and issue should be fixed_


bigcat88:
image

Updated docs

bigcat88:
Second/third issue(s):

When I press "Deploy and Enable" I see that progress go and I made a page refresh(F5).
Button "Remove" appeared near the progress what I think is should not happen?

Also after that UI stopped on ending progress and ExAPP in UI do not reach "Enabled" state.

andrey18106 (approved):

PR #245, opened 2024-02-27
Save user specified daemon host when using manual install

MB-Finski:
At least for manual install, currently, a custom host is not saved as the front-end overwrites it which is quite confusing. 'host.docker.internal' is not resolvable under a Linux install of docker by default so it might make sense to not enforce this as a default setting in any case.


bigcat88 (approved):
Yep, we missed this, this was intended only for our temporary developing.

Thanks you, good catch!

PR #278, opened 2024-04-23
fix(occ): incorrect type of field in ExAppOccCommand

andrey18106:
Should fix incorrect type handling by different DBMS during ExApp occ command registration.

### DIFF
--- a/lib/Db/Console/ExAppOccCommand.php
+++ b/lib/Db/Console/ExAppOccCommand.php
@@ -15,7 +15,7 @@ use OCP\AppFramework\Db\Entity;
  * @method string getAppid()
  * @method string getName()
  * @method string getDescription()
- * @method bool getHidden()
+ * @method int getHidden()
  * @method array getArguments()
  * @method array getOptions()
  * @method array getUsages()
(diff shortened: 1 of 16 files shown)


lib/Service/ExAppOccService.php
```diff
@@ -124,7 +124,7 @@ public function __construct(
 			protected function configure() {
 				$this->setName($this->occCommand->getName());
 				$this->setDescription($this->occCommand->getDescription());
-				$this->setHidden($this->occCommand->getHidden());
+				$this->setHidden(filter_var($this->occCommand->getHidden(), FILTER_VALIDATE_BOOLEAN));
```
bigcat88, line 127: what is the purpose of it? it is already `int` on input and `int` in the database

lib/Db/Console/ExAppOccCommand.php
```diff
@@ -87,7 +87,7 @@ public function jsonSerialize(): array {
 			'appid' => $this->getAppid(),
 			'name' => $this->getName(),
 			'description' => $this->getDescription(),
-			'hidden' => $this->getHidden(),
+			'hidden' => (bool) $this->getHidden(),
```
bigcat88, line 90: this commit already marked it as a `int` why to convert to `bool`?

lib/Service/ExAppOccService.php
bigcat88: never mind, this is fine

bigcat88 (approved):

PR #294, opened 2024-05-17
feat: Add task processing API

provokateurin:


lib/Migration/Version2700Date20240515092246.php
```diff
@@ -0,0 +1,76 @@
...
+				'notnull' => true,
+				'length' => 64,
+			]);
+			$table->addColumn('task_type', Types::STRING, [
+				'notnull' => true,
+				'length' => 64,
+			]);
+
+			$table->setPrimaryKey(['id']);
+			$table->addUniqueIndex(['app_id', 'name'], 'task_processing_idx');
```
bigcat88, line 71: should be here the `task_type` added as well to the unique index, as this is how registering of task types in new API goes?
or will they have different `name` always?

example: 

appid=some_provider
name=llm_llama_v3
task_type=summar
task_type=topics_extract
task_type=some_other_tasktype

provokateurin: Makes sense :+1:

bigcat88:
@provokateurin can you fix psalm-ci please, so we can merge it?

bigcat88 (approved):

PR #301, opened 2024-06-07
Add removeData parameter to ExApp Remove UI

### LINKED ISSUE #297: Add option to remove volume on ExApp removal

Currently, ExApps management is limited to the default narrowed number of actions (Install, enable/disable, remove).

It would be useful to extend ExApps management page, in particular the "Remove" action to allow selecting the option whether to remove ExApp volume (docker volume) or not (simple flag already available in API).

Add checkbox somewhere on ExApps management page (in sidebar) to allow administrator to remove ExApp and it's volume from the daemon.

_No response_

### Additional context

image

edward-ly:
Closes #297.

### DIFF
--- a/src/store/apps.js
+++ b/src/store/apps.js
@@ -295,10 +295,10 @@ const actions = {
 		}).catch((error) => context.commit('API_FAILURE', { appId, error }))
 	},
 
-	uninstallApp(context, { appId }) {
+	uninstallApp(context, { appId, removeData }) {
 		return api.requireAdmin().then((response) => {
 			context.commit('startLoading', appId)
-			return api.get(generateUrl(`/apps/app_api/apps/uninstall/${appId}`))
+			return api.get(generateUrl(`/apps/app_api/apps/uninstall/${appId}/${removeData}`))
 				.then((response) => {
 					context.commit('stopLoading', appId)
 					context.commit('uninstallApp', appId)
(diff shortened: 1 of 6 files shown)


edward-ly:
I noticed while making the PR that if there is more than one ExApp installed and disabled, the value of the checkbox persists when switching between apps in the list, and resets after closing and reopening the sidebar or leaving the page. I just want to confirm if this behavior is intended (or, if not, what the correct behavior should be).

src/store/apps.js
```diff
@@ -295,10 +295,10 @@ const actions = {
 		}).catch((error) => context.commit('API_FAILURE', { appId, error }))
 	},
 
-	uninstallApp(context, { appId }) {
+	uninstallApp(context, { appId, removeData }) {
 		return api.requireAdmin().then((response) => {
 			context.commit('startLoading', appId)
-			return api.get(generateUrl(`/apps/app_api/apps/uninstall/${appId}`))
+			return api.get(generateUrl(`/apps/app_api/apps/uninstall/${appId}/${removeData}`))
```
bigcat88, line 301: if there is no strong reason to make it a path parameter, please, change it to query parameter

bigcat88:
> I noticed while making the PR that if there is more than one ExApp installed and disabled, the value of the checkbox persists when switching between apps in the list, and resets after closing and reopening the sidebar or leaving the page. I just want to confirm if this behavior is intended (or, if not, what the correct behavior should be).

it should always be in an unchecked state when switching between apps in the list (even if it was checked, then the focus was switched to another ExApp and switched back)

src/store/apps.js
edward-ly: As in the URL string should be `/apps/app_api/apps/uninstall/${appId}?removeData=${removeData}` instead?
That's the only other way I can think of; I don't quite understand how the `generateUrl` function is able to add query parameters otherwise (I thought it was the [`params`](https://nextcloud-libraries.github.io/nextcloud-router/functions/generateUrl.html) argument at first, but that turned out to be wrong).

edward-ly:
> it should always be in an unchecked state when switching between apps in the list (even if it was checked, then the focus was switched to another ExApp and switched back)

Now added in fdf70df.

bigcat88 (approved):

bigcat88:
Thanks for the pull request @edward-ly

PR #305, opened 2024-06-12
UIProxy: fixed incorrect the Domain of `cookie`

bigcat88:
Currently cookies always rejected when ExApp is installed in DSP with `HTTPS` - this is now fixed.


kyteinsky (approved):

kyteinsky:
oops, didn't see it was a draft. Still looked good for yolo.

PR #323, opened 2024-07-10
optimization: system flag and ex_app_users removal

andrey18106:
The system flag and extra table for that `ex_app_users` is removed to optimize database usage and simplify the system logic.


lib/Service/AppAPIService.php
```diff
@@ -430,9 +411,6 @@ private function finalizeRequestToNC(string $userId, IRequest $request, int $isS
 			$this->userSession->setUser(null);
 		}
 		$this->session->set('app_api', true);
-		if ($isSystemApp === 1) {
-			$this->session->set('app_api_system', true);
```
bigcat88, line 434: this should be present, until NC29 support drop

bigcat88 (approved):
Need additional manual testing/check with providing `info.xml` without `system` flag - to see if all is ok,- but looks good to me.
Release with this anyway will be on the next week, so we will have plenty time to test it locally and see that it is not breaking anything during development of ExApps

PR #324, opened 2024-07-11
feat(TaskProcessing): Allow defining custom task types

provokateurin:


andrey18106 (approved):
Didn't test, but LGTM

marcelklehr (commented):
Conceptually, I would say the new task type should be independent from the provider. E.g. one app could register multiple providers for the same custom task type. Also, one app could want to register only a task type for a task to be provided by other apps.

bigcat88:
> Conceptually, I would say the new task type should be independent from the provider. E.g. one app could register multiple providers for the same custom task type.

Will this implementation be enough at this stage?
This does not require creating a new table, and is simpler from a API point of view, IMHO.

bigcat88:
```python3

    nc.providers.task_processing.register(
        "sdxl_lighting",
        "SDXL Lighting",
        "visionatrix:sdxl_lighting",
        {
            "id": "visionatrix:sdxl_lighting",
            "name": "SDXL Lighting",
            "description": "Fast Image Generation",
            "input_shape": {
                "prompt": {
                    "name": "prompt",
                    "description": "Prompt",
                    "type": 1,  # Text
                }
            },
            "output_shape": {
                "image": {
                    "name": "image",
                    "description": "Image",
                    "type": 2,  # Image
                }
            }
        }
    )
```

Description of idea: if one ExApp provides multiple custom_task_types for each custom task_type it calls "Register Provider" (as anyway "Optional Shapes" are binded to Provider struct)

_I am still in process of reviewing and learning how to use our new API.._

bigcat88 (changes requested):
PR is fine for me, but I can't make it work with the Assistant app.

No further actions is needed(until someone want to help find to which repo the problem belongs), current idea is to take a much deeper look at this at Monday/Tuesday and find the reason.

bigcat88 (approved):
Ok, it was a bug in the Assistant app - it was fixed 20 minutes ago thanks to Julien for fast reacting.

marcelklehr:
> Will this implementation be enough at this stage?

I'm uncertain on this. think the goal of ExApps was to have stable APIs. Having this one in a half baked state does not contribute to that goal. On the other hand it may be possible to extend this implementation in the future. But then I don't see when we would do that. IMO it would be good to have this implemented properly, but I defer to @julien-nc

PR #326, opened 2024-07-12
feat(PublicFunctions): add getExApp method

andrey18106:


lib/PublicFunctions.php
```diff
@@ -101,4 +101,25 @@ public function asyncExAppRequestWithUserInit(
 		}
 		return $this->service->requestToExAppAsync($exApp, $route, $userId, $method, $params, $options, $request);
 	}
+
+	/**
+	 * Get basic ExApp info by appid
+	 *
+	 * @param string $appId
+	 *
+	 * @return array ExApp info (appid, version, name, enabled) or error message
```
bigcat88, line 110: or `null`

bigcat88 (approved):

bigcat88:
I am testing this right now with new Nextcloud Server Webhook system and it looks fine to me.
We can merge this, imho

PR #334, opened 2024-07-22
fix: preserve original Authentication to be swapped later in HaProxy

andrey18106:
Ref: https://github.com/cloud-py-api/docker-socket-proxy/pull/33


bigcat88:
swapping should only takes place when HaProxy is used and there should not be swap when "raw" socket is used.

bigcat88 (approved):
Now this applies only for remote installations(when using https and request to ExApp come throught DSP), regarding changes for docker-socket-proxy is here:

https://github.com/cloud-py-api/docker-socket-proxy/pull/33

PR #327, opened 2024-07-16
ExApp routes (public/user/admin) support

andrey18106:
This PR introduces new (**mandatory if you use the ExApp proxy**) registration of the routes that are allowed to call on ExApp via AppAPI ExApp proxy.

In your ExApp ``info.xml`` - ``external-app`` add the ``routes`` declaration, e.g.:

```
<routes>
	<route>
		<url>.*</url>
		<verb>GET,POST,PUT,DELETE</verb>
		<access_level>USER</access_level>
		<headers_to_exclude>[]</headers_to_exclude>
	</route>
</routes>
```
allows all routes, or more specific:

```
<routes>
	<route>
		<url>img\/.*</url>
		<verb>GET</verb>
		<access_level>USER</access_level>
		<headers_to_exclude>[]</headers_to_exclude>
	</route>
	<route>
		<url>js\/.*</url>
		<verb>GET</verb>
		<access_level>USER</access_level>
		<headers_to_exclude>[]</headers_to_exclude>
	</route>
	<route>
		<url>css\/.*</url>
		<verb>GET</verb>
		<access_level>USER</access_level>
		<headers_to_exclude>[]</headers_to_exclude>
	</route>
	<route>
		<url>api\/.*</url>
		<verb>GET,POST,PUT,DELETE</verb>
		<access_level>USER</access_level>
		<headers_to_exclude>[]</headers_to_exclude>
	</route>
</routes>
```

PR #335, opened 2024-07-23
Revert "drop support of Nextcloud 27 (#322)"

andrey18106:
This reverts commit ca4c14c7.

We need this to keep the #327 included in NC 27.

We'll drop it's support after AppAPI 3.0.0.


bigcat88 (approved):
looks good

PR #338, opened 2024-07-28
Add possibility to edit Deploy daemon configuration

### LINKED ISSUE #332: Add possibility to edit Deploy daemon configuration

It would be more convenient to have an "Edit Deploy daemon" feature.

It's actually the same modal as for "Register Deploy daemon" but slightly adjusted with already filled data from the target Deploy daemon to edit. In this way we keep all the logic of the registration checks.
The backend part also needs the endpoint to update the edited Deploy daemon.

For each Deploy daemon list item add context menu option "Edit", which opens a modal with data of selected daemon configuration, similar to "Register modal", but without templates, and with adjusted warnings logic.

_No response_

### Additional context

There are some fields, the changes of which are reflected only for newly installed ExApps (Docker container re-create is required

Fields that should be able to be edited without warning:

* `Display Name`, `Daemon Host`, `HaProxyPassword` (in the deployment config)

With warning:

* `Nextcloud Url`, `Network` (in deployment config), `Compute Device`  (in deployment config)

image

vstelmakh:
Resolves #332

image


andrey18106 (changes requested):
Is there any reason for manual changes to `l10n/*.js` files? They are being updated automatically via Transifex (https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#adding-translations).

vstelmakh:
> Is there any reason for manual changes to `l10n/*.js` files? They are being updated automatically via Transifex (https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#adding-translations).

Thank you for bringing this to my attention. I had no idea about the update process for the translation files in this project. My intention was simply to improve the translations related to the change for the languages that I understand.

Should I revert the changes I made to the translation files?

Please let me know how I can best proceed to ensure everything remains consistent with the project's workflow.

bigcat88:
> Should I revert the changes I made to the translation files?

Yes, please, revert the translation changes.

We try to avoid making changes to multiple topics in a single pull request to make them easier to review and keep the commit and discussion history cleaner.

andrey18106 (approved):
LGTM, tested, thanks for the improvements and some refactoring.

bigcat88 (approved):

PR #348, opened 2024-08-03
fix: disable caching if APCu detected

bigcat88:
We do not disable caches in other places as in CLI only `/ex_apps` cache is used.

Linked issue: https://github.com/cloud-py-api/app_api/issues/291


bigcat88:
It would also be a good idea to first create a separate PR with tests that will fail and merge it, and then rebase this PR and make sure that this fix works - this will help you avoid falling into the same hole in the future.

andrey18106 (approved):
LGTM, tested.

PR #351, opened 2024-08-05
feat(ExAppProxy): add `X-Origin-IP` header for rate-limiting purposes…

andrey18106:
… for ExApp


bigcat88 (approved):
Cool! Now ExApps that needs Public Routes can protect itself with RateLimiting

PR #359, opened 2024-08-08
feat: Add enum and default value support in task processing

kyteinsky:
All related PRs:
https://github.com/cloud-py-api/nc_py_api/pull/284
https://github.com/cloud-py-api/app_api/pull/359
https://github.com/nextcloud/translate2/pull/11


lib/Service/ProvidersAI/TaskProcessingService.php
```diff
@@ -81,11 +84,18 @@ public function registerTaskProcessingProvider(
 			$taskProcessingProvider = null;
 		}
 		try {
+			if ($customTaskType !== null && $taskType !== $customTaskType['id']) {
+				throw new Exception('Task type and custom task type must be the same if custom task type is provided');
+			} elseif ($customTaskType === null && $provider['task_type'] === null) {
+				throw new Exception('Task type must be provided if custom task type is not provided');
+			}
+
```
marcelklehr, line 92: I think it would be good to have more validation here to give early feedback to the exapp that the registration failed. Otherwise if the passed array is wrong, it will return ok, but when we want to use the provider it will fail without the exapp getting wind of it.

bigcat88 (changes requested):
We have time to change this, since NC30 is not published yet - let's make everything simple and clear in AppAPI too.

I propose to transmit/receive everything in the structure that Anupam introduced(`array $provider`), without duplicating the same information in `POST` parameters as separate variables.

`$name`, `$displayName`, `$taskType` at leat should be removed as a parameters, maybe `$customTaskType` too.

composer.lock
andrey18106, line 1: Maybe we should update composer files in a separate PR if it's needed?

bigcat88: yes ,you are right, but let's make an exception this time for this PR

bigcat88 (approved):
_I'd published nc_py_api_

@marcelklehr will you have time to test this tommorow?

Anupam told me that he tested and it was working, from my point of view all is looking fine.

But as I understand this all is a breaking change in API and we need to publish all ExApp related to this in one day.

marcelklehr:
> @marcelklehr will you have time to test this tommorow?

Yepm will do

marcelklehr:
> But as I understand this all is a breaking change in API and we need to publish all ExApp related to this in one day(or at least make commit to not break "Daily").

Right. Personally, I'm ok if AI on daily doesn't work for a bit, but if we can avoid it that'd be good. That would be app_api 4.0.0 then? Affected ex apps apps of mine is only llm2, I think

lib/Service/ProvidersAI/TaskProcessingService.php
```diff
@@ -68,24 +70,106 @@ public function getExAppTaskProcessingProvider(string $appId, string $name): ?Ta
...
+		if (!isset($provider['id']) || !is_string($provider['id'])) {
+			throw new Exception('"id" key must be a string');
+		}
+		if (!isset($provider['name']) || !is_string($provider['name'])) {
+			throw new Exception('"name" key must be a string');
+		}
+		if (!isset($provider['task_type']) || !is_string($provider['task_type'])) {
+			throw new Exception('"task_type" key must be a string');
+		}
+		if (!isset($provider['expected_runtime']) || !is_int($provider['expected_runtime'])) {
```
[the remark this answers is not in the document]
marcelklehr: Or we just leave it as is.

marcelklehr:
Tested and works with llm2

marcelklehr (approved):

PR #366, opened 2024-08-22
Proxy: check all routes until we find allowing route

bigcat88:
Yes, it changes the algorithm, but otherwise it does not work in PGSQL in production.

And so it is a little simpler, since the order of routes in `info.xml` is not important


andrey18106 (approved):
LGTM

PR #369, opened 2024-08-22
call `throttler->resetDelay` if only when delay present

bigcat88:
This will require **one less network request** for each successful request from ExApp


andrey18106 (approved):
👍️

PR #381, opened 2024-09-06
fix(cli): do not remove persistent storage by default in ExApp unregister command

andrey18106:
Do not remove ExApp docker volume by default, deprecate old option `keep-data`, add new `rm-data` to remove ExApp data.


bigcat88 (approved):
👍

PR #391, opened 2024-09-19
fix(webhooks_unregister): add missing method_exists check

andrey18106:
Add additional safety check if mapper's method exists


bigcat88:
In what cases may a method not exist if this method was originally added from start of NC 30.0?

Developer RC versions can be ignored, as they are already in the past.

bigcat88 (changes requested):
awaits clarification, what does it really fixes

PR #394, opened 2024-09-25
Feat/daemon selection

### LINKED ISSUE #276: Daemon selection for ExApp installation

It would be useful to be able to select on which daemon to install ExApp, not only to default one.

Add option in the UI to select on which daemon (or daemons cluster) to install ExApp.

_No response_

### Additional context

Also this might be relevant and required in case of scaled ExApps to be able to install ExApp on multiple daemons, but this requires more advanced daemons configuration (groups or clusters of daemons) and possibility to select on which group of daemons to setup ExApps.

janepie:
Resolves #276 

I had to prolong the waiting time for the registering as (at least on my setup) it took 15-20 sec (also when done per occ command directly, I checked)

Did not comment the Vue parts as you don't seem to use comments there. If I should add comments just let me know!


andrey18106:
It seems there are some syntax errors in new components (missing prop, error prop type check) in console while opening daemon selection - I don't see the list of daemons to choose from (I have one manual-install and one docker-install daemon).

janepie:
Then you shouldn't see the list as it's only one valid daemon. Can you check with two docker-install?

janepie:
Should be fixed, but something in my NC instance seems broken right now and I cannot validate. I will try to fix it and check, feel free to check yourself or wait :)

janepie:
Works!

bigcat88 (approved):
Thank you for putting effort on this, great job.

bigcat88:
now it’s up to you, Andrey, when we will combine this and whether we will make adjustments on our part.

PR #404, opened 2024-09-26
[stable30] chore(ci): update tests CI for stable30

andrey18106:
Update tests CI for stable30 branch only


andrey18106:
Deploy tests in Docker fail is expected as the app is not force enabled (stable30 currently requires NC 30.0.1), CI should be green after server stable30 version is >=30.0.1

PR #436, opened 2024-10-31
Fix: check healthcheck and heartbeat after ExApp container start

bigcat88:
After the previous PR(#430) which added stopping application containers feature when shutting down, as @andrey18106  correctly noted during our 1:1 conversation:

we cannot rely on the application inside the docker container to be ready to process the request at the moment the docker container is turned on.

This PR introduces handling of it by checking health-check and heartbeat.


andrey18106 (approved):
LGTM 👍

PR #447, opened 2024-11-18
(fix): do not expose haproxy password to UI

oleksandr-nc:
1. Since we need a password of at least 12 characters, we use `dummySecret123` instead of `dummySecret`
2. `checkDaemonConnection` expects to get the full Deploy Daemon configuration from the UI, so we get the password from the database if we get `dummySecret123` from the UI and the haproxy password is present in the database.
3. I added comments everywhere in the code

### DIFF
--- a/lib/Controller/DaemonConfigController.php
+++ b/lib/Controller/DaemonConfigController.php
@@ -59,11 +59,37 @@ class DaemonConfigController extends ApiController {
 	#[PasswordConfirmationRequired]
 	public function updateDaemonConfig(string $name, array $daemonConfigParams): Response {
 		$daemonConfig = $this->daemonConfigService->getDaemonConfigByName($name);
+
+		// Safely check if "haproxy_password" exists before accessing it
+		$haproxyPassword = $daemonConfigParams['deployConfig']['haproxy_password'] ?? null;
+
+		// Restore the original password if "dummySecret123" is provided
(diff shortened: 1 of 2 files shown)


lib/Controller/DaemonConfigController.php
```diff
@@ -59,11 +59,37 @@ public function registerDaemonConfig(array $daemonConfigParams, bool $defaultDae
 	#[PasswordConfirmationRequired]
 	public function updateDaemonConfig(string $name, array $daemonConfigParams): Response {
 		$daemonConfig = $this->daemonConfigService->getDaemonConfigByName($name);
+
+		// Safely check if "haproxy_password" exists before accessing it
+		$haproxyPassword = $daemonConfigParams['deployConfig']['haproxy_password'] ?? null;
+
+		// Restore the original password if "dummySecret123" is provided
+		if ($haproxyPassword === 'dummySecret123') {
+			$daemonConfigParams['deployConfig']['haproxy_password'] = $daemonConfig->getDeployConfig()['haproxy_password'] ?? "";
```
kyteinsky, line 68: `deploy_config` instead of `deployConfig`

oleksandr-nc: Keen eye, thanks for your attention!

kyteinsky (approved):
:rocket:

PR #451, opened 2024-11-20
(fix): Encrypt sensitive values(haproxy_password) in the DB

oleksandr-nc:
Subj. Added a migration for encryption, it needs to be executed manually for the master branch (I didn't increase the application version, because this branch will only be released with Nextcloud 31):

`sudo -u www-data php occ migrations:execute app_api 5000Date20241120135411`

For backports, we will increase the application version, and in Nextcloud 30 or 29 nothing will need to be done manually.


lib/Migration/Version5000Date20241120135411.php
```diff
@@ -0,0 +1,59 @@
...
+			if (!empty($deployConfig['haproxy_password'])) {
+				$deployConfig['haproxy_password'] = $this->crypto->encrypt($deployConfig['haproxy_password']);
+				$encodedDeployConfig = json_encode($deployConfig);
+				$qbUpdate = $this->connection->getQueryBuilder();
+				$qbUpdate->update('ex_apps_daemons')
+					->set('deploy_config', $qbUpdate->createNamedParameter($encodedDeployConfig))
+					->where(
+						$qbUpdate->expr()->eq('id', $qbUpdate->createNamedParameter($row['id']))
+					);
+				$qbUpdate->executeStatement();
```
julien-nc, line 53: This query could be generated once with parameters and reused in each iteration. But since there will likely not be a lot of daemon configs, it will surely not impact performance.

kyteinsky (approved):

lib/Migration/Version5000Date20241120135411.php
```diff
@@ -0,0 +1,59 @@
...
+	 * @param array $options
+	 *
+	 * @return null|ISchemaWrapper
+	 */
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options) {
+		// encrypt "haproxy_password" in the "ex_apps_daemons" table
+		$qbSelect = $this->connection->getQueryBuilder();
+		$qbSelect->select(['id', 'deploy_config'])
+			->from('ex_apps_daemons')
+			->where(1);
```
marcelklehr, line 38: This breaks postgres due to type mismatch.

PR #465, opened 2024-12-05
README - fix two doc links: daemon, ext. apps

meonkeys:


oleksandr-nc (approved):
Thank you for spotting this

Issue #468, opened 2024-12-10
Authenticated ciphertext could not be decoded


tobiasKaminsky:
### Describe the bug

```
{
  "reqId": "ObUipdnIDrRtHVTcJDyU",
  "level": 3,
  "time": "2024-12-10T09:51:59+00:00",
  "remoteAddr": "2a02:8071:64e0:7bc0:2f86:72c1:6a64:1cb5",
  "user": "admin",
  "app": "index",
  "method": "GET",
  "url": "/settings/apps",
  "message": "Authenticated ciphertext could not be decoded.",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64; rv:133.0) Gecko/20100101 Firefox/133.0",
  "version": "31.0.0.6",
  "exception": {
    "Exception": "Exception",
    "Message": "Authenticated ciphertext could not be decoded.",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/html/lib/private/Security/Crypto.php",
        "line": 102,
        "function": "decryptWithoutSecret",
        "class": "OC\\Security\\Crypto",
        "type": "->",
        "args": [
          "*** sensitive parameters replaced ***"
        ]
      },
      {
        "file": "/var/www/html/apps/app_api/lib/DeployActions/DockerActions.php",
        "line": 626,
        "function": "decrypt",
        "class": "OC\\Security\\Crypto",
        "type": "->",
        "args": [
          "*** sensitive parameters replaced ***"
        ]
      },
      {
        "file": "/var/www/html/apps/app_api/lib/Service/ExAppsPageService.php",
        "line": 46,
        "function": "initGuzzleClient",
        "class": "OCA\\AppAPI\\DeployActions\\DockerActions",
        "type": "->"
      },
      {
        "file": "/var/www/html/apps/settings/lib/Controller/AppSettingsController.php",
        "line": 97,
        "function": "provideAppApiState",
        "class": "OCA\\AppAPI\\Service\\ExAppsPageService",
        "type": "->"
      },
      {
        "file": "/var/www/html/lib/private/AppFramework/Http/Dispatcher.php",
        "line": 200,
        "function": "viewApps",
        "class": "OCA\\Settings\\Controller\\AppSettingsController",
        "type": "->"
      },
      {
        "file": "/var/www/html/lib/private/AppFramework/Http/Dispatcher.php",
        "line": 114,
        "function": "executeController",
        "class": "OC\\AppFramework\\Http\\Dispatcher",
[...]

PR #469, opened 2024-12-10
chore: bump version to trigger new migration

andrey18106:
No version bump was added with new migration added (https://github.com/nextcloud/app_api/pull/451).

Resolves: https://github.com/nextcloud/app_api/issues/468


andrey18106:
Ah, no, we shouldn't bump here version to keep it in sync with NC minor and patch versions.

Issue #472, opened 2024-12-17
[Bug]: Warning message in log (defaultDaemonConfig) when navigating to "Applications" tab and submenu


ozgurkazancci:
Got the "**Invalid defaultDaemonConfig data provided to provideInitialState by settings**" warning issue under Nextcloud 30.0.4, always filling Nextcloud log with such warning;

```
{
  "reqId": "WARRV71KCzCb08PKkEY4",
  "level": 2,
  "time": "2024-12-17T06:03:38+00:00",
  "remoteAddr": "xx.xx.xx",
  "user": "admin",
  "app": "no app in context",
  "method": "GET",
  "url": "/settings/apps",
  "message": "Invalid defaultDaemonConfig data provided to provideInitialState by settings",
  "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36",
  "version": "30.0.4.1",
  "data": [],
  "id": "67611464470d8"
}
```

P.S.: Disabling "app_api" stops producing such warning.

PR #473, opened 2024-12-17
fix(initial-state): correct defaultDaemonConfig initial state

andrey18106:
There is was missing check if no deploy daemons configured, null value produced warning logs.

Resolves: https://github.com/nextcloud/app_api/issues/472


oleksandr-nc (approved):
LGTM, I hope the frontend will handle the missing `defaultDaemonConfig` correctly

*Don't forget to increase the version in a separate PR, for the `stable30` branch, so that it is the same minor as the server*

Issue #474, opened 2024-12-20
Updating an external app must not automatically enable it


nickvergessen:
## PHP apps
- Have a list of enabled and disabled apps
- Run `occ app:update --all`
- List of enabled and disabled apps is the same, apart from the version

## AppAPI apps
- Have a list of enabled and disabled apps
- Run `occ app_api:app:update --all`
- All apps that were updated are now also enabled 💥 

The difference is unexpected.
AppAPI apps that were disabled before should stay disabled

PR #478, opened 2024-12-26
fix(occ): correct `app_api:app:update` enabled state of ExApps, introduce helper options

andrey18106:
Resolves: #474 

This PR corrects AppAPI `occ app_api:app:update` command logic in the following way:

1. By default disabled ExApps are not updated. Use option `--include-disabled` to update disabled ExApps too.
2. During update ExApp is enabled to perform initialization step, if it was disabled before update - by default it will be disabled after update.

### DIFF
--- a/lib/Command/ExApp/Update.php
+++ b/lib/Command/ExApp/Update.php
@@ -50,10 +50,9 @@ class Update extends Command {
 		$this->addOption('force-scopes', null, InputOption::VALUE_NONE, 'Force new ExApp scopes approval[deprecated]');
 		$this->addOption('wait-finish', null, InputOption::VALUE_NONE, 'Wait until finish');
 		$this->addOption('silent', null, InputOption::VALUE_NONE, 'Do not print to console');
-		$this->addOption('all', null, InputOption::VALUE_NONE, 'Update all enabled and updatable apps');
+		$this->addOption('all', null, InputOption::VALUE_NONE, 'Updates all enabled and updatable apps');
 		$this->addOption('showonly', null, InputOption::VALUE_NONE, 'Additional flag for "--all" to only show all updatable apps');
-		$this->addOption('include-disabled', null, InputOption::VALUE_NONE, 'Update also disabled apps (use --enable-after-update to enable ExApp after update)');
-		$this->addOption('enable-after-update', null, InputOption::VALUE_NONE, 'Enable ExApp after update if it was disabled before. Use with --include-disabled');
+		$this->addOption('include-disabled', null, InputOption::VALUE_NONE, 'Additional flag for "--all" to also update disabled apps');
 	}
 
 	protected function execute(InputInterface $input, OutputInterface $output): int {
@@ -156,9 +155,8 @@ class Update extends Command {
 		$exApp->setStatus($status);
 		$this->exAppService->updateExApp($exApp, ['status']);
 
-		$wasEnabled = false;
+		$wasEnabled = $exApp->getEnabled();
 		if ($exApp->getEnabled()) {
-			$wasEnabled = true;
 			if ($this->service->disableExApp($exApp)) {
 				$this->logger->info(sprintf('ExApp %s successfully disabled.', $appId));
 				if ($outputConsole) {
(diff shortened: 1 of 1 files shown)


lib/Command/ExApp/Update.php
```diff
@@ -50,8 +50,10 @@ protected function configure(): void {
 		$this->addOption('force-scopes', null, InputOption::VALUE_NONE, 'Force new ExApp scopes approval[deprecated]');
 		$this->addOption('wait-finish', null, InputOption::VALUE_NONE, 'Wait until finish');
 		$this->addOption('silent', null, InputOption::VALUE_NONE, 'Do not print to console');
-		$this->addOption('all', null, InputOption::VALUE_NONE, 'Update all updatable apps');
+		$this->addOption('all', null, InputOption::VALUE_NONE, 'Update all enabled and updatable apps');
```
oleksandr-nc, line 53: Update**s** ...

lib/Command/ExApp/Update.php
```diff
@@ -50,8 +50,10 @@ protected function configure(): void {
...
 		$this->addOption('showonly', null, InputOption::VALUE_NONE, 'Additional flag for "--all" to only show all updatable apps');
+		$this->addOption('include-disabled', null, InputOption::VALUE_NONE, 'Update also disabled apps (use --enable-after-update to enable ExApp after update)');
```
oleksandr-nc, line 55: Also update**s** disabled apps (use --enable-after-update to enable **an** ExApp after **the** update)

lib/Command/ExApp/Update.php
```diff
@@ -50,8 +50,10 @@ protected function configure(): void {
...
+		$this->addOption('enable-after-update', null, InputOption::VALUE_NONE, 'Enable ExApp after update if it was disabled before. Use with --include-disabled');
```
oleksandr-nc, line 56: Enable**s** **an** ExApp after **the** update if it was disabled before. Use with --include-disabled

Question: 

    What is the purpose of this option? When might it be useful?
    Perhaps it shouldn't be added, to avoid complicating the logic too much?

*If the decision to add it is correct, it should be documented clearly, explaining when, why, and for what it can be used.*

lib/Command/ExApp/Update.php
```diff
@@ -145,7 +156,9 @@ private function updateExApp(InputInterface $input, OutputInterface $output, str
 		$exApp->setStatus($status);
 		$this->exAppService->updateExApp($exApp, ['status']);
 
+		$wasEnabled = false;
```
oleksandr-nc, line 159:
```php
$wasEnabled = $exApp->getEnabled();
if ($wasEnabled) {
   ...
```

oleksandr-nc (changes requested):

lib/Command/ExApp/Update.php
andrey18106: Right, this is redundant, removed.

lib/Command/ExApp/Update.php
```diff
@@ -145,6 +155,7 @@ private function updateExApp(InputInterface $input, OutputInterface $output, str
 		$exApp->setStatus($status);
 		$this->exAppService->updateExApp($exApp, ['status']);
 
+		$wasEnabled = $exApp->getEnabled();
 		if ($exApp->getEnabled()) {
```
bigcat88, line 159: if ($wasEnabled) {

lib/Command/ExApp/Update.php
```diff
@@ -110,6 +111,15 @@ private function updateExApp(InputInterface $input, OutputInterface $output, str
 			return 1;
 		}
 
+		$includeDisabledApps = $input->getOption('include-disabled');
+		if ($input->getOption('all') && !$exApp->getEnabled() && !$includeDisabledApps) {
+			$this->logger->error(sprintf('ExApp %s is disabled. Update skipped (use --include-disabled to update disabled apps).', $appId));
```
oleksandr-nc, line 116: should be info `info` level, imho

oleksandr-nc (approved):

PR #482, opened 2025-01-03
(feat): Allow any docker network for HTTPS(remote) install type

oleksandr-nc:
This PR adds support for ExApp installations when the protocol type is `https` and the network daemon type is not equal to `host` value.

*Additionally, deployment tests have been modified to include more information to help detect problems faster.*

Required for applications like ContextChat or Flow that contain a PgSQL database that used default port that may conflict with each other without network isolation.

_Note: this PR does not include a modification of the interface part where these restrictions will be removed. This should be presented in a separate PR, you can check it through console commands by creating a daemon or by editing the value in the DB manually._


lib/DeployActions/DockerActions.php
```diff
@@ -301,6 +301,23 @@ public function createContainer(string $dockerUrl, string $imageId, array $param
...
+			$containerParams['ExposedPorts'] = [
+				sprintf('%d/tcp', $exAppMainPort) => (object) [],
+				sprintf('%d/udp', $exAppMainPort) => (object) [],
+			];
+			$containerParams['HostConfig']['PortBindings'] = [
+				sprintf('%d/tcp', $exAppMainPort) => [
+					['HostPort' => (string)$exAppMainPort, 'HostIp' => '127.0.0.1']
+				],
+				sprintf('%d/udp', $exAppMainPort) => [
+					['HostPort' => (string)$exAppMainPort, 'HostIp' => '127.0.0.1']
```
kyteinsky, line 316: we can use `$deployConfig['additional_options']['OVERRIDE_APP_HOST'];` here and `127.0.0.1` as the fallback IP if the override env var is not provided.
Since `OVERRIDE_APP_HOST` can be a domain name, I'm not sure how should we handle that since the DNS of the remote server would be used to resolve it.

oleksandr-nc: After a 1:1 discussion, it was decided not to add OVERRIDE_APP_HOST for now.

PR #488, opened 2025-01-09
fix: add more space to taskprocessing columns

kyteinsky:


oleksandr-nc:
versioning is now as follows:

main -> 32.0.0-dev.`x`
stable31 -> 5.0.`x` (there was an error here initially, so that they wouldn't add the dev.0 prefix)
stable30 -> 4.0.`x`

If we need to apply a migration for `main`, `stable31` or `stable30` we will increase the value of X

_Note: CI may be red now, we will need to reconfigure CI after the fork, I will take care of that._

andrey18106 (approved):

oleksandr-nc (approved):

PR #505, opened 2025-02-07
enh: add exapps info endpoint for HaRP

kyteinsky:
Related PRs:

- Test-Deploy: https://github.com/nextcloud/test-deploy/pull/11
- App-Skeleton-Python: https://github.com/nextcloud/app-skeleton-python/pull/34
- Julius Docker-Dev: https://github.com/juliusknorr/nextcloud-docker-dev/pull/382
- NC-Py-API: https://github.com/cloud-py-api/nc_py_api/pull/338


lib/Controller/HarpController.php
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	#[PublicPage]
+	#[NoCSRFRequired]
+	public function getExAppsMetadata(): DataResponse {
+		// todo: temporary solution to use app config
+		$harpKey = $this->appConfig->getValueString(Application::APP_ID, 'harp_shared_key');
+		$headerHarpKey = $this->request->getHeader('HARP-SHARED-KEY');
+		if ($headerHarpKey === '' || $headerHarpKey !== $harpKey) {
+			return new DataResponse(['message' => 'Harp shared key is not valid'], Http::STATUS_UNAUTHORIZED);
```
bigcat88, line 53: Better from start to add `throttler`:

https://github.com/nextcloud/app_api/blob/982c7813ad32696d79e598c68b919f2a3ac62424/lib/Service/AppAPIService.php#L284-L289

lib/Command/Daemon/RegisterDaemon.php
```diff
@@ -32,17 +32,16 @@ protected function configure(): void {
...
 
-		// todo: add docs
-		$this->addArgument('name', InputArgument::REQUIRED);
+		$this->addArgument('name', InputArgument::REQUIRED, 'Unique deploy daemon name');
 		$this->addArgument('display-name', InputArgument::REQUIRED);
-		$this->addArgument('accepts-deploy-id', InputArgument::REQUIRED);
-		$this->addArgument('protocol', InputArgument::REQUIRED);
-		$this->addArgument('host', InputArgument::REQUIRED);
+		$this->addArgument('accepts-deploy-id', InputArgument::REQUIRED, 'The deployment method that the daemon accepts. Can be "manual-install" or "docker-install". "docker-install" is for Docker Socket Proxy and HaRP.');
+		$this->addArgument('protocol', InputArgument::REQUIRED, 'The protocol used to connect to the daemon. Can be "http" or "https".');
```
oleksandr-nc, line 38: will it always be `http` for `HaRP` and `manual-install`?

kyteinsky: don't think so, manual installs can be behind proxies or support https directly
same for harp, can be a proxy's address. The protocol is not used anywhere else for harp I think.

oleksandr-nc (approved):
approve in advance - since it won't be backported and the sooner we merge it, the more bugs we can find before NC32 release

lib/Controller/HarpController.php
```diff
@@ -0,0 +1,166 @@
...
+		private readonly ICrypto             $crypto,
+		private readonly ?string             $userId,
+	) {
+		parent::__construct(Application::APP_ID, $request);
+
+		$this->request = $request;
+	}
+
+	private function validateHarpSharedKey(string $appId): bool {
+		// todo: cache
```
oleksandr-nc, line 50: will be good to take DeployConfig from the ExApp object. It is present there.

 _to introduce the new cache is not the way to go, as we should decrease number of request from AppAPI to the DB/Redis and not increase them._

kyteinsky: yeah agreed
thanks for the advice, it is indeed much cleaner this way!

andrey18106:
@kyteinsky Hey, just tested works fine, except a few errors in Admin settings UI:

1. Enable HaRP in Register modal not working as expected (wrong template name)
2. Edit Deploy daemon not working on old daemons (they don't have `deploy_config.harp` key) - maybe post-process prepare daemons deploy config?

<details>

image

</details>

kyteinsky:
Thanks @andrey18106, please check if those fixed the issues.

lib/Controller/HarpController.php
```diff
@@ -0,0 +1,165 @@
...
+		private readonly DaemonConfigService $daemonConfigService,
+		private readonly ICrypto             $crypto,
+		private readonly ?string             $userId,
+	) {
+		parent::__construct(Application::APP_ID, $request);
+
+		$this->request = $request;
+	}
+
+	private function validateHarpSharedKey(string $appId): bool {
```
bigcat88, line 49: should accept `ExApp` instead of `string`, imho.

In `getExAppMetadata` then first will be call to `getExApp` and this will be passed to `validateHarpSharedKey`

This will help to call `getExApp` only once instead of two times for the  fetching ExApp info from HaRP(one network call takes 3ms-20ms).

In `getUserInfo` just needed to add new the `getExApp` call, nothing will be changed except ~5 more lines of code total.

kyteinsky: nice!

Issue #498, opened 2025-01-18
Questions about the 'deploy configurations'


DecaTec:
I'm referring here to the docs 'Deploy configurations': https://nextcloud.github.io/app_api/DeployConfigurations.html

Questions in **bold**.

---

When I got it right, the first (and easiest) configuration is 'NC and Docker on the same host' (see [here](https://nextcloud.github.io/app_api/DeployConfigurations.html#nc-docker-on-the-same-host)):
- The Docker Socket Proxy is not needed in this case.
- Direct communications with Docker via `/var/docker/docker.sock`

I've configured this exactly as mentioned in the docs:

> Daemon host: /var/run/docker.sock
>HTTPS checkbox: not supported using docker socket
>Network: host
>HaProxy password: not supported using raw docker socket, should be empty

When testing the connection, this fails and I get following error in the NC logs:
```
ConnectException
cURL error 7: Failed to connect to localhost port 80 after 0 ms: Couldn't connect to server (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://localhost/v1.41/_ping
Could not connect to Docker daemon
```
**Question 1: Why it tries to connect to localhost on port 80? Shouldn't it use the Docker socket directly (without HTTP communication)?**

**Question 2: Am I missing something here? Is the documentation lacking an important piece of information?**

---

Using the Docker  Socket Proxy, see "Suggested way to communicate with Docker via [Docker Socket Proxy container](https://github.com/cloud-py-api/docker-socket-proxy)" it works then using method A or B.

**Question 3: What are the advantages/disadvantages between "direct communication via `/var/run/docker.sock`" and using the Docker Socket Proxy? Is there one recommended way when NC, Docker and ext. apps should run on the same system?**

---

Thanks in advance!

oleksandr-nc:
> Question 1: Why it tries to connect to localhost on port 80? Shouldn't it use the Docker socket directly (without HTTP communication)?

Probably it have no access to the file(does te webuser can access docker socket file?), as **file_exists** function fails.

```php

public function buildDockerUrl(DaemonConfig $daemonConfig): string {
		if (file_exists($daemonConfig->getHost())) {
			return 'http://localhost';
		}
		return $daemonConfig->getProtocol() . '://' . $daemonConfig->getHost();
	}
```

> Question 3: What are the advantages/disadvantages between "direct communication via `/var/run/docker.sock`" and using the Docker Socket Proxy?

[...]

PR #510, opened 2025-02-12
Improve error logging

R0Wi:
* Check local socket by assuming it starts with a slash
* Improve error logging for ping and deploy

Addresses #498 

Log messages now look like this when doint a ping via UI (if either the docker socket is not writable or the file path doesn't exist)

*socket doesn't exist*:
socket-does-not-exist

*socket is not writable*:
socket-not-writable

Also the frontend dialog for the deployment now shows

*deployment error for local socket*:
deploy-result-err

*deployment error for nextcloud DSP*:
deploy-result-err-dsp

Unfortunately I didn't find a way to get rid of the slightly misleading error message that - in case the socket is not usable - curl will always complain about not being able to connect tho `http://localhost/`, even though it's not using any http under the hood but trying to access the UNIX socket directly. When using curl in socket mode ( `CURLOPT_UNIX_SOCKET_PATH`), it will still need some `http`-address for the actual `GET` call. See also https://github.com/curl/curl/issues/1338

### DIFF
--- a/lib/DeployActions/DockerActions.php
+++ b/lib/DeployActions/DockerActions.php
@@ -37,6 +37,7 @@ class DockerActions implements IDeployActions {
 
 	private Client $guzzleClient;
 	private bool $useSocket = false;  # for `pullImage` function, to detect can be stream used or not.
+	private string $socketAddress;
 
 	public function __construct(
 		private readonly LoggerInterface           $logger,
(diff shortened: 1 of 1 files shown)


oleksandr-nc:
@kyteinsky or @andrey18106 - second review?

*note: we will backport this to the NC31/NC30*

lib/DeployActions/DockerActions.php
```diff
@@ -823,21 +826,20 @@ public function healthcheckContainer(string $containerId, DaemonConfig $daemonCo
 	}
 
 	public function buildDockerUrl(DaemonConfig $daemonConfig): string {
-		if (file_exists($daemonConfig->getHost())) {
-			return 'http://localhost';
-		}
-		return $daemonConfig->getProtocol() . '://' . $daemonConfig->getHost();
+		// When using local socket, we the curl URL needs to be set to http://localhost
```
oleksandr-nc, line 829: except for a small typo in the comment, at first glance everything looks good in this PR

oleksandr-nc (approved):

andrey18106 (approved):
LGTM

lib/DeployActions/DockerActions.php
R0Wi: Will fix the typo with my next PR 👍

Issue #517, opened 2025-02-17
[bug] Issue when trying to install external app via appstore


R0Wi:
### Describe the bug

When trying to install an ExApp via Nextcloud Appstore (UI), the app gets stuck and then fails

I added a few debug log statements and found this:

```json
{
   "reqId":"RxuK0gxgmrpPF0sscwUP",
   "level":1,
   "time":"2025-02-16T13:21:07+00:00",
   "remoteAddr":"172.25.0.1",
   "user":"admin",
   "app":"app_api",
   "method":"POST",
   "url":"/apps/app_api/apps/enable/workflow_ocr_backend",
   "message":"Calling occ(directory=null): 'app_api:app:register' '--silent' 'workflow_ocr_backend' '' '' --no-ansi --no-warnings",
   "userAgent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/132.0.0.0 Safari/537.36",
   "version":"31.0.0.15",
   "data":{
      "app":"app_api"
   },
   "id":"67b39c0b146aa"
}

{
   "reqId":"RxuK0gxgmrpPF0sscwUP",
   "level":0,
   "time":"2025-02-16T13:21:07+00:00",
   "remoteAddr":"172.25.0.1",
   "user":"admin",
   "app":"app_api",
   "method":"POST",
   "url":"/apps/app_api/apps/enable/workflow_ocr_backend",
   "message":"occ error output: \n                                                                               \n  Too many arguments to \"app_api:app:register\" command, expected arguments \"a  \n  ppid\" \"daemon-config-name\".                                                  \n                                                                               \n\napp_api:app:register [--force-scopes] [--info-xml INFO-XML] [--json-info JSON-INFO] [--wait-finish] [--silent] [--test-deploy-mode] [--env [ENV]] [--mount [MOUNT]] [--] <appid> [<daemon-config-name>]\n\n",
   "userAgent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/132.0.0.0 Safari/537.36",
   "version":"31.0.0.15",
   "data":{
      "app":"app_api"
   },
   "id":"67b39c0b146a1"
}
```

To me it seems that the `occ` command, which is called under the hood, fails, because the parameters are not being setup correctly (note the "empty" parameters)

### Steps/Code to Reproduce

1. Install NC31
2. Install AppApi and Deploy daemon
[...]

PR #518, opened 2025-02-17
Fix parameter preparation for occ command

R0Wi:
* Fix for #517
* Filters empty `occ` commandline params
* Adds improved logging in case this ever happens again (it was pretty hard to find the bug because after the `occ` command has been sent, nothing could be found in the logs)

**Notes**

* I was only able to reproduce this on NC31 (because I didn't find any NC32 compatible apps in the store, yet). So the base for this fix is `stable31`
* I took the liberty of defaulting `$this->config->getSystemValue('apps_paths')` with `[]`, because I found some errors in my logs when trying to uninstall an app (something like "foreach needs an array, string given ..."). This happens if `app_paths` is not set. `getSystemValue` will return an "empty" string in that case. 
* If possible, a testcase should be added for this. Maybe with any deployed app which is updated regularly? Note that the problem did not occur when doing the "test install"

### DIFF
--- a/lib/Service/AppAPIService.php
+++ b/lib/Service/AppAPIService.php
@@ -428,8 +428,9 @@ class AppAPIService {
 	 */
 	public function runOccCommand(string $command): bool {
 		$args = array_map(function ($arg) {
-			return escapeshellarg($arg);
+			return $arg === '' ? null : escapeshellarg($arg);
 		}, explode(' ', $command));
+		$args = array_filter($args, fn ($arg) => $arg !== null);
 		$args[] = '--no-ansi --no-warnings';
 		return $this->runOccCommandInternal($args);
 	}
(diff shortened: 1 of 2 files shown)


lib/Service/AppAPIService.php
```diff
@@ -428,8 +428,9 @@ public function dispatchExAppInitInternal(ExApp $exApp): void {
 	 */
 	public function runOccCommand(string $command): bool {
 		$args = array_map(function ($arg) {
-			return escapeshellarg($arg);
+			return empty($arg) ? null : escapeshellarg($arg);
```
oleksandr-nc, line 431: empty('0') will return `True`. Maybe better to compare with `return $arg === ''` ?

oleksandr-nc (changes requested):
Was on vacation, sorry for the delay in replying.

And why is PR made in `stable31` branch? Maybe it would be better to make it in main and backport it?

oleksandr-nc:
@R0Wi good time of day. Are you willing to rebase this PR on `main` branch or should we do it ourselves?

R0Wi:
Hi @oleksandr-nc, sorry for the delayed answer. Unfortunately I've been out sick last week.

Happy to incorporate your review findings tonight 👍

R0Wi:
@oleksandr-nc rebased on `main` and incorporated your feedback. Needs to be backported to `stable31` ...

kyteinsky (approved):

oleksandr-nc (approved):

oleksandr-nc:
> @oleksandr-nc rebased on `main` and incorporated your feedback. Needs to be backported to `stable31` ...

Thank you

PR #532, opened 2025-03-25
(chore): removed deprecated Speech, Tanslate and TextProcessing AI providers

oleksandr-nc:
PR to speed up AppAPI for Nextcloud 32.

Each of the providers always added **one extra request to the cache/DB** for any request to the NC instance, which slows down the entire instance (_a guess_).

I completely cut them all out, they were deprecated in Nextcloud 30(?) (and a universal TaskProcessing provider was introduced).


marcelklehr:
It's a bit worrisome that we deprected only last year and are now removing this already, but since it's only in AppAPI, we could get away with it... Not sure if we need approval from Andy or somebody else

julien-nc:
Andy is fine with it as long as we have a TaskProcessing replacement for all the old API's providers implemented in exApps. And we have that, right?

andrey18106 (approved):

oleksandr-nc:
> Well the translation task type in task processing is not backward compatible with the old translate API, but other than that we're good.

But the old translate API has not been used anywhere since NC31?

marcelklehr:
> But the old translate API has not been used anywhere since NC31?

Yeah, we're good :)

marcelklehr (approved):

PR #542, opened 2025-04-04
(fix): perfomance - load ExApps menu items only when event is triggered

oleksandr-nc:
Results for 5k requests:

OCS call, AppAPI enabled(before PR):
```
Maximum request time: 0.01429 seconds
Minimum request time: 0.01301 seconds
Average request time: 0.01395 seconds
```

OCS call, AppAPI enabled(after PR):
```
Maximum request time: 0.01383 seconds
Minimum request time: 0.01281 seconds
Average request time: 0.01347 seconds (3.5% performance boost compared to old code)
```

_For frontend calls the speed is almost the same, slight boost with less then 1%, when two ExApps define their icons._

### Why this code is writtent without usual initilizing in the `__construct()`?

```php
$menuEntries = Server::get(TopMenuService::class)->getExAppMenuEntries();
...
/** @var IGroupManager $groupManager */
$groupManager = Server::get(IGroupManager::class);
/** @var INavigationManager $navigationManager */
$navigationManager = Server::get(INavigationManager::class);
/** @var IURLGenerator $urlGenerator */
$urlGenerator = Server::get(IURLGenerator::class);
/** @var IFactory $l10nFactory */
$l10nFactory = Server::get(IFactory::class);
```		

This is much faster, as this does not initialize all this classes for each request, but only initialize them for those requests where we should do the actual work.
_First version of this PR was written in a nice way(with `private readonly ...` in constructor), but after measurments the code was completly rewritten._

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -16,6 +16,7 @@ use OCA\AppAPI\Listener\DeclarativeSettings\RegisterDeclarativeSettingsListener;
 use OCA\AppAPI\Listener\DeclarativeSettings\SetValueListener;
 use OCA\AppAPI\Listener\FileEventsListener;
 use OCA\AppAPI\Listener\LoadFilesPluginListener;
+use OCA\AppAPI\Listener\LoadMenuEntriesListener;
 use OCA\AppAPI\Listener\SabrePluginAuthInitListener;
 use OCA\AppAPI\Middleware\AppAPIAuthMiddleware;
 use OCA\AppAPI\Middleware\ExAppUIL10NMiddleware;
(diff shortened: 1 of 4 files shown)


oleksandr-nc:
question - will we backport this to the NC31 or not?

kyteinsky (approved):
:rocket:

> question - will we backport this to the NC31 or not?

okay to do imo

oleksandr-nc:
> I don't get that. You only give the class name to the register listener. The listener is not initialized until the event is actually triggered.
> If that is no longer the case it's a serious bug.

Ok, I will recheck that.

oleksandr-nc:
I forced-pushed a slightly modified version.

You are right, the class is not initialized until there is an event, so we can put `TopMenuService` in the constructor, because it will always be used first during an event and it makes no difference where it is located.

After taking measurements when there are ExApps but none of them define a menu item, the code without IGroupManager, INavigationManager, IURLGenerator, IFactory in the constructor still works faster (or it is a measurement error as difference value is not so big).

I would like to leave them where they are, if this does not cause any problems that I do not know about - but if it is needed, I can push a version where they are initialized in the constructor.

Issue #530, opened 2025-03-18
Making development easier: support deploying from local/custom Docker registries


oleksandr-nc:
Currently, AppAPI determines the Docker image source exclusively from the `appinfo/info.xml` file, structured as follows:

```xml
<external-app>
  <docker-install>
    <registry>ghcr.io</registry>
    <image>nextcloud/flow</image>
    <image-tag>1.1.0</image-tag>
  </docker-install>
</external-app>
```

This approach is inconvenient for local development, especially in the final stages of testing, because each change requires pushing the image to GitHub and subsequently downloading it again for verification. This process is particularly inefficient for larger images (e.g., PyTorch-based images that easily exceed 5 GB).

Proposed Enhancement:

On the admin settings page (`https://nextcloud.local/settings/admin/app_api`), we add a new item called “Override Docker Registries” to the Deploy Daemons drop-down menu.
  > This option should appear only for daemons whose deployment type (accepts-deploy-id) is set to "docker-install".

When selecting this new option, a simple configuration window should appear, allowing administrators to define registry override rules—for example, replacing the registry specified in <docker-install> with a custom/local Docker registry. These overrides should be stored within the `deploy_config` JSON field of the `oc_ex_apps_daemons` table.

Expected Behavior:

When deploying an external app, AppAPI should first check if a registry override is configured. If present, AppAPI must use the custom-defined registry instead of the registry specified in `appinfo/info.xml`.

Benefits:
  * Simplifies the development process by enabling deployments directly from local Docker storage.
  * Enables customization of Docker registries, beneficial in Enterprise environments or other scenarios with specialized Docker Images storage requirements.

PR #554, opened 2025-04-14
feat: custom Docker registries support (only for docker-install type)

andrey18106:
This PR adds possibility to configure custom docker registry mappings.

Resolves: https://github.com/nextcloud/app_api/issues/530

In addition to UI there are 3 simple CLI commands added to manage daemon's custom registry mappings:

- `occ app_api:daemon:registry:add`
- `occ app_api:daemon:registry:remove`
- `occ app_api:daemon:registry:list`

Registry mapping stored in Deploy daemon's `deploy_config.registries` (`{from: string, to: string}[]`).
The mapping target `to="local"` disables image pulling step, ExApp image name is not changed and retrieved from `info.xml` as usually, assuming it was already pulled manually or built locally with the same name.
In other cases image_src (registry domain) is replaced (e.g. `ghcr.io/repository/image_name -> docker.io/repository/image_name`.

<details>

image

image

</details>


lib/Db/ExAppMapper.php
```diff
@@ -200,6 +200,9 @@ public function registerExAppRoutes(ExApp $exApp, array $routes): int {
 			if (isset($route['bruteforce_protection']) && is_string($route['bruteforce_protection'])) {
 				$route['bruteforce_protection'] = json_decode($route['bruteforce_protection'], false);
 			}
+			if (!isset($route['headers_to_exclude'])) {
```
oleksandr-nc, line 203: this should be in a separate PR, as we need to backport it.

oleksandr-nc (approved):

oleksandr-nc:
And `#[PasswordConfirmationRequired]` is missing in the endpoints..

kyteinsky (commented):
looks nice!
But what do you think of making it just an override value that replaces any and all registries? Like value "local" could replace docker.io and ghcr.io, the `from` field would not be required. Right now we would have to check the info.xml file for the registry that the apps are using to put in the `from` value.
Let me know if I misunderstood something.

oleksandr-nc:
> But what do you think of making it just an override value that replaces any and all registries?

For some situations this will be not optimal. For example in some org, there will be only banned(or they will think that it is insecure) `ghcr.io` and  the`docker.io` registry will be fine for them.

Then they can make a private registry only for one public registry and mirror only ExApps from it and leave other untouched.

kyteinsky (approved):
:rocket:

PR #555, opened 2025-04-16
feat: added ability to disable FRP for HaRP<-->ExApp communication. 

oleksandr-nc:
Will be useful for AIO and future support of the Kubernetes.

So this advanced option make HaRP work more like old DSP deploy method and have all those problems with dificulty of setup. This option  is disabled by default, as usual home users do not need it at all.
Differences between old DSP and this option:

1. Requests still comes through HaRP, so Nextcloud just informs HaRP to where reroute the request, to the FRP tunnel or is that direct communication.
2. `appid` will be resolved to the IP address  inside HaRP and not on the Nextcloud side, which should me much better.

### DIFF
--- a/lib/Command/Daemon/RegisterDaemon.php
+++ b/lib/Command/Daemon/RegisterDaemon.php
@@ -48,6 +48,7 @@ class RegisterDaemon extends Command {
 		$this->addOption('harp_frp_address', null, InputOption::VALUE_REQUIRED, '[host]:[port] of the HaRP FRP server, default host is same as HaRP host and port is 8782');
 		$this->addOption('harp_shared_key', null, InputOption::VALUE_REQUIRED, 'HaRP shared key for secure communication between HaRP and AppAPI');
 		$this->addOption('harp_docker_socket_port', null, InputOption::VALUE_REQUIRED, '\'remotePort\' of the FRP client of the remote docker socket proxy. There is one included in the harp container so this can be skipped for default setups.', '24000');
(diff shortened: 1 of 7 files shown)


andrey18106 (approved):

lib/Command/Daemon/RegisterDaemon.php
```diff
@@ -48,6 +48,7 @@ protected function configure(): void {
 		$this->addOption('harp_frp_address', null, InputOption::VALUE_REQUIRED, '[host]:[port] of the HaRP FRP server, default host is same as HaRP host and port is 8782');
 		$this->addOption('harp_shared_key', null, InputOption::VALUE_REQUIRED, 'HaRP shared key for secure communication between HaRP and AppAPI');
 		$this->addOption('harp_docker_socket_port', null, InputOption::VALUE_REQUIRED, '\'remotePort\' of the FRP client of the remote docker socket proxy. There is one included in the harp container so this can be skipped for default setups.', '24000');
+		$this->addOption('harp_exapp_direct', null, InputOption::VALUE_NONE, 'Flag for the advanced setups only. Disables the FRP tunnel between ExApps and HaRP.');
```
kyteinsky, line 51: this can be displayed in `daemon:list` too

kyteinsky (approved):

lib/Command/Daemon/RegisterDaemon.php
oleksandr-nc: maybe later, this is too specific option, imho.

PR #567, opened 2025-04-17
fix: add HTTPS toggle back in HaRP

kyteinsky:


kyteinsky:
this would need additional fixups because of special handling of "protocol"
to be continued later

Issue #562, opened 2025-04-17
Daemon name with spaces leads to not working ExApps


oleksandr-nc:
### Describe the bug

{"reqId":"IPRW4hYFtNIOA2obJyXv","level":1,"time":"2025-04-17T07:01:50+00:00","remoteAddr":"XXXX","user":"admin","app":"app_api","method":"POST","url":"/apps/app_api/daemons/CPU%20-%20S20/test_deploy","message":"Calling occ(directory=null): 'app_api:app:register' '--silent' 'test-deploy' 'CPU' '-' 'S20' '--info-xml' 'https://raw.githubusercontent.com/nextcloud/test-deploy/main/appinfo/info.xml' '--test-deploy-mode' --no-ansi --no-warnings","userAgent":"Mozilla/5.0 (X11; Linux x86_64; rv:136.0) Gecko/20100101 Firefox/136.0","version":"32.0.0.0","data":{"app":"app_api"}}
{"reqId":"IPRW4hYFtNIOA2obJyXv","level":3,"time":"2025-04-17T07:01:50+00:00","remoteAddr":"XXXX","user":"admin","app":"app_api","method":"POST","url":"/apps/app_api/daemons/CPU%20-%20S20/test_deploy","message":"Error executing occ command. Return code: 1, stdout: , stderr: \n \n Too many arguments to "app_api:app:register" command, expected arguments "a \n ppid" "daemon-config-name". \n \n\napp_api:app:register [--force-scopes] [--info-xml INFO-XML] [--json-info JSON-INFO] [--wait-finish] [--silent] [--test-deploy-mode] [--env [ENV]] [--mount [MOUNT]] [--] <appid> [<daemon-config-name>]\n\n","userAgent":"Mozilla/5.0 (X11; Linux x86_64; rv:136.0) Gecko/20100101 Firefox/136.0","version":"32.0.0.0","data":{"app":"app_api"}}

### Steps/Code to Reproduce

Just register daemon with space in `Name`

### Expected Results

We should block such names(ids) from registering or correctly escape and process their names.

### Actual Results

Test deploy should work!

### Setup configuration

All versions. Found on Nextcloud 31.0.4.RC1

PR #575, opened 2025-04-29
fix(DeployDaemon): support for daemon names containing spaces

oleksandr-nc:
Fixes: #562 

_If backport will not fail, this should be backported, imho._


kyteinsky (approved):
:rocket:

PR #585, opened 2025-05-26
feat(declarativeSettings): support sensitive Declarative settings values encryption

andrey18106:
Related to: https://github.com/nextcloud/server/pull/53121

Implement sensitive Declarative settings encryption for ExApps.


oleksandr-nc:
In context agent we use `appconfig_ex.get_value` to read declarative setting values(_from fast search by repo sources_, @janepie knows for sure).

I guess, that decryption for sensitive values should be added there as well?

janepie:
> In context agent we use `appconfig_ex.get_value` to read declarative setting values(_from fast search by repo sources_, @janepie knows for sure).
> 
> I guess, that decryption for sensitive values should be added there as well?

Yes, also `set_value` right now

lib/Service/ExAppConfigService.php
```diff
@@ -43,6 +54,14 @@ public function getAppConfigValues(string $appId, array $configKeys): ?array {
 
 	public function setAppConfigValue(string $appId, string $configKey, mixed $configValue, ?int $sensitive = null): ?ExAppConfig {
```
oleksandr-nc, line 55: Tests failures are reproducable locally. 

They fails cause previous behaviour was to return the data that are stored in the settings, and with this PR it returns encrypted.
My proposal to keep the behaviour as it was before and to return from this function decrypted `configvalue`.

lib/Migration/Version032002Date20250527174907.php
```diff
@@ -0,0 +1,89 @@
...
+	 *
+	 * @return null|ISchemaWrapper
+	 */
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options) {
+		// encrypt appconfig_ex values that have sensitive flag set
+
+		$qbSelect = $this->connection->getQueryBuilder();
+		$qbSelect->select(['id', 'configvalue'])
+			->from('appconfig_ex')
+			->where($qbSelect->expr()->eq('sensitive', $qbSelect->createNamedParameter(1)));
```
kyteinsky, line 65: the type param for `createNamedParameter` may be required here since the default is string

lib/Migration/Version032002Date20250527174907.php
```diff
@@ -0,0 +1,89 @@
...
+		while ($row = $req->fetch()) {
+			$configValue = $row['configvalue'];
+			if (!empty($configValue)) {
+				try {
+					$encryptedValue = $this->crypto->encrypt($configValue);
+					$qbUpdate = $this->connection->getQueryBuilder();
+					$qbUpdate->update('appconfig_ex')
+						->set('configvalue', $qbUpdate->createNamedParameter($encryptedValue))
+						->where(
+							$qbUpdate->expr()->eq('id', $qbUpdate->createNamedParameter($row['id']))
```
kyteinsky, line 77: type param here too

lib/Service/ExAppConfigService.php
```diff
@@ -65,6 +84,9 @@ public function setAppConfigValue(string $appId, string $configKey, mixed $confi
 				return null;
 			}
 		}
+		if ($sensitive) {
+			$appConfigEx->setConfigvalue($configValue);
```
kyteinsky, line 88: A comment would be nice here, something like "Setting the plain text value in the object for local usage".

lib/Service/ExAppConfigService.php
```diff
@@ -43,20 +54,28 @@ public function getAppConfigValues(string $appId, array $configKeys): ?array {
 
 	public function setAppConfigValue(string $appId, string $configKey, mixed $configValue, ?int $sensitive = null): ?ExAppConfig {
 		$appConfigEx = $this->getAppConfig($appId, $configKey);
+		if ($sensitive) {
+			try {
+				$encryptedValue = $this->crypto->encrypt($configValue);
```
kyteinsky, line 59: empty string check can be done here in the `if` otherwise empty string values cannot be set for sensitive fields. It would throw an exception and return.

lib/Service/ExAppPreferenceService.php
```diff
@@ -25,48 +26,73 @@ class ExAppPreferenceService {
...
-	public function setUserConfigValue(string $userId, string $appId, string $configKey, mixed $configValue) {
+	public function setUserConfigValue(string $userId, string $appId, string $configKey, mixed $configValue, ?int $sensitive = null): ?ExAppPreference {
 		try {
 			$exAppPreference = $this->mapper->findByUserIdAppIdKey($userId, $appId, $configKey);
 		} catch (DoesNotExistException|MultipleObjectsReturnedException|Exception) {
 			$exAppPreference = null;
 		}
+		if ($sensitive) {
+			try {
+				$encryptedValue = $this->crypto->encrypt($configValue);
```
kyteinsky, line 41: same empty string check here too

lib/Service/ExAppPreferenceService.php
```diff
@@ -25,48 +26,73 @@ class ExAppPreferenceService {
...
 				}
-				return $exAppPreference;
 			} catch (Exception $e) {
 				$this->logger->error('Error while updating config value: ' . $e->getMessage(), ['exception' => $e]);
 				return null;
 			}
 		}
+		if ($sensitive) {
+			$exAppPreference->setConfigvalue($configValue);
+		}
```
kyteinsky, line 77: A similar comment for using the plain text value for the sensitive/encrypted field.

kyteinsky (approved):
:rocket:

Issue #587, opened 2025-06-02
Why is https not supported in manual-install?


BerengarWLehr:
I'm running one ExApp for our Universitys NC-Instance in our own K8s-Cluster. Adding  a deployment daemon into this setup would need additional fragmentation due to necessary separation.
I absolutely can see that it may be easier for others to let a deamon handle deploying containers but I know how to run my setup and would like to keep it that way.
What bothers me is that I can not use httpS between NC and ExAPP (https://github.com/nextcloud/app_api/blob/db0f559e8ac6c0b333dd3e25405e1ebd1e629a3c/lib/Command/Daemon/RegisterDaemon.php#L76). Is there some explanation why? I could imagine that the docker daemon will roll out it's own certificates but we have our own certificate infrastructure that we can use for subdomains and so forth.
Maybe someone already has suggestions what could be done to allow a safe and encrypted communication between both ends when deployed manually?

BerengarWLehr:
I have not considered using VPN for this problem as until now I cannot find a reason why httpS would not be supported. I thought about just commenting out the lines and "just try it" but wanted to check in first as I assume there are is reason behind this.

oleksandr-nc:
The installation manual was originally made for the development of ExApps, that's why it doesn't have `https`

oleksandr-nc:
I guess we can close this?

oleksandr-nc:
Feel free to reopen if needed, or better to open a discussion instead of an issue.

PR #593, opened 2025-06-05
 Remove https prevention for manual-install

BerengarWLehr:
Following up on the discussion https://github.com/nextcloud/app_api/issues/587 I'd suggest that the prevention of using https for the manual-install Daemon should be removed as it doesn't seem to have any inherit reason.

There are indeed real world use cases in which manual-install is an equal strategy for deploying containers for ExApps. There are sophisticated ways to let NC run the ExAppContainer but sometimes running it manually is just another way to reach Rome.

PS: This is the 2nd corrected version of https://github.com/nextcloud/app_api/pull/591 m(

### DIFF
--- a/lib/Command/Daemon/RegisterDaemon.php
+++ b/lib/Command/Daemon/RegisterDaemon.php
@@ -72,10 +72,6 @@ class RegisterDaemon extends Command {
 			$output->writeln('Value error: The protocol must be `http` or `https`.');
 			return 1;
 		}
-		if ($acceptsDeployId === 'manual-install' && $protocol !== 'http') {
-			$output->writeln('Value error: Manual-install daemon supports only `http` protocol.');
-			return 1;
-		}
 		if ($isHarp && !$input->getOption('harp_shared_key')) {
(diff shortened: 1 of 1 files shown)


oleksandr-nc:
Hello. What did corrected in this version? I just do not get why you had close two other PR's..

BerengarWLehr:
@oleksandr-nc in the first PR I deleted the wrong lines (sloppy me deleted the if clause three lines above) in the 2nd I forgot to sign it of and couldn't for the love of god figure out how to repair this.

oleksandr-nc (approved):

andrey18106 (approved):

oleksandr-nc:
Thanks for the PR. 
But I am not sure that we can backport this without adjusting documentation and testing that it works.

@BerengarWLehr if you can test this on Nextcloud 32 (to which current `main` branch belongs) and say that if it works for your case will be very nice. Of course if you want. After that we can accept PR to the documentation and backport this to Nextcloud 31 if needed.

PR #608, opened 2025-07-03
fix: streaming

lukasdotcom:
Streaming did not work for me when using ex app routes and downloading a large file through them. Not sure if this is a correct or valid way to fix it, but it does allow downloading large files.

### DIFF
--- a/lib/Controller/ExAppProxyController.php
+++ b/lib/Controller/ExAppProxyController.php
@@ -48,7 +48,7 @@ class ExAppProxyController extends Controller {
 		parent::__construct(Application::APP_ID, $request);
 	}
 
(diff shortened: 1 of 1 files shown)


bigcat88:
At first look:

The problem that `stream` can break `createProxyResponse` with current changes.

We need at least check for `$isHTML` in the code(`ExAppGet`,...) and if it `true` then we will skip using `stream`.

To keep code clean we need to pass flag `isHTML` in the `createProxyResponse` to not duplicate code and not do double check for `isHTML`.

oleksandr-nc (approved):
Thank you very much for not only finding the bug, but also making a fix.

PR #612, opened 2025-07-11
feat(new-route): added endpoint to call NC getAbsoluteURL

oleksandr-nc:
This is needed to resolve the following issue: https://github.com/nextcloud/context_agent/issues/46

The endpoint will be backported to Nextcloud 31.

You can test it with the following snippet:

```
print(nc.ocs("GET", f"{nc._session.ae_url}/info/nextcloud_url/absolute", json={"url": "some_url"}))
```

P.S. After we merge this, it might be a good idea to add a method to nc_py_api to allow calling it more cleanly.


kyteinsky (approved):
:rocket:

Issue #618, opened 2025-07-26
Post update 30.0.13 app_api failures


corsairius:
### Describe the bug

I don't know if this is a red herring, but this appeared to occur after an update to the latest minor version of Nextcloud.  This appeared to occur both in richdocuments and llm2, and whisper2 getting a 400:

```
[PHP] Warning: foreach() argument must be of type array|object, string given at /usr/share/nginx/nextcloud/apps/app_api/lib/Fetcher/ExAppArchiveFetcher.php#114
	GET /apps/app_api/apps/uninstall/stt_whisper2?removeData=false
	from 192.168.64.1 by jcollins at Jul 23, 2025, 11:50:13 PM
[app_api] Error: Error during registering ExApp llm2.
	from ? by -- at Jul 23, 2025, 11:43:18 PM
```
and
```

{"reqId":"CAEH6GzDdd0pXktugJz0","level":2,"time":"2025-07-24T04:43:18+00:00","remoteAddr":"","user":"--","app":"PHP","method":"","url":"--","message":"Undefined array key \"name\" at /usr/share/nginx/nextcloud/apps/app_api/lib/Service/ExAppService.php#86","userAgent":"--","version":"30.0.13.1","data":{"app":"PHP"},"id":"6881bbe7a13ce"}

{"reqId":"wdAjsb6Qpgx1uRqDIbJl","level":3,"time":"2025-07-24T13:42:02+00:00","remoteAddr":"xxx","user":"xxx","app":"richdocuments","method":"GET","url":"/ocs/v2.php/apps/files/api/v1/templates","message":"Client error: `POST https://xxx/apps/richdocumentscode/proxy.php?filter=contentcontrol` resulted in a `400 Bad Request` response:\n<html><body>\n<h1>Socket proxy error</h1>\n<p>Error: The param should be &#039;status&#039; or &#039;req=...&#039;, but is (truncated...)\n","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36 Edg/138.0.0.0","version":"30.0.13.1","data":{"app":"richdocuments"},"id":"6882382cbce92"}


[PHP] Warning: foreach() argument must be of type array|object, string given at /usr/share/nginx/nextcloud/apps/app_api/lib/Fetcher/ExAppArchiveFetcher.php#114
[...]

PR #620, opened 2025-07-30
[stable30] fix(exAppArchiveFetcher): correct apps_path handling

andrey18106:
Add fallback to default `/apps/` folder.

Related to: #618

### DIFF
--- a/lib/Fetcher/ExAppArchiveFetcher.php
+++ b/lib/Fetcher/ExAppArchiveFetcher.php
@@ -90,7 +90,7 @@ class ExAppArchiveFetcher {
 	}
 
 	public function getExAppFolder(string $appId): ?string {
-		$appsPaths = $this->config->getSystemValue('apps_paths');
+		$appsPaths = $this->config->getSystemValue('apps_paths', []);
 		$existingPath = '';
 		foreach ($appsPaths as $appPath) {
 			if ($appPath['writable'] && file_exists($appPath['path'] . '/' . $appId)) {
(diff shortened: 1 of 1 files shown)


lib/Fetcher/ExAppArchiveFetcher.php
```diff
@@ -107,15 +107,27 @@ public function getExAppFolder(string $appId): ?string {
...
 	public function removeExAppFolder(string $appId): void {
-		foreach ($this->config->getSystemValue('apps_paths') as $appPath) {
-			if ($appPath['writable']) {
-				if (file_exists($appPath['path'] . '/' . $appId)) {
-					$this->rmdirr($appPath['path'] . '/' . $appId);
-				}
+		$appsPaths = $this->config->getSystemValue('apps_paths', []);
+		if (empty($appsPaths)) {
+			// check if the default apps folder exists
+			$defaultAppsFolder = \OC::$SERVERROOT . '/apps/' . $appId;
```
oleksandr-nc, line 122: let's name the var `defaultAppFolder` or `defaultExAppFolder` to not insult someone who will read that we removing folder with all applications

lib/Fetcher/ExAppArchiveFetcher.php
```diff
@@ -107,15 +107,27 @@ public function getExAppFolder(string $appId): ?string {
...
+			if (is_dir($defaultAppsFolder)) {
+				$this->rmdirr($defaultAppsFolder);
+			}
+			return;
+		}
+		foreach ($this->config->getSystemValue('apps_paths', []) as $appPath) {
```
oleksandr-nc, line 128: can we just use the already defined `$appsPaths` variable here?

PR #621, opened 2025-07-30
fix: for transfer encoding causing problems

lukasdotcom:
Returning a ProxyResponse with a transfer-encoding header causes the response to be cut off and nothing to be returned this removes the transfer-encoding header when it is set to chunking. MCP servers respond with that header breaking the proxy.


oleksandr-nc:
> Should this be backported to stable30 also?

yes, why not.

PR #625, opened 2025-08-05
fix(exAppArchiveFetcher): correct apps_path handling

andrey18106:
Add fallback to default /apps/ folder.
Related to: https://github.com/nextcloud/app_api/issues/618

### DIFF
--- a/lib/Fetcher/ExAppArchiveFetcher.php
+++ b/lib/Fetcher/ExAppArchiveFetcher.php
@@ -95,7 +95,7 @@ class ExAppArchiveFetcher {
 	}
 
 	public function getExAppFolder(string $appId): ?string {
-		$appsPaths = $this->config->getSystemValue('apps_paths');
+		$appsPaths = $this->config->getSystemValue('apps_paths', []);
 		$existingPath = '';
 		foreach ($appsPaths as $appPath) {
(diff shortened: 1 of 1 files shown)


kyteinsky:
From https://github.com/nextcloud/app_api/issues/618#issuecomment-3124541393 and https://docs.nextcloud.com/server/latest/admin_manual/apps_management.html#using-custom-app-directories,
it seems like `apps_path` is always supposed to be an array but it wasn't in the linked issue so it was a config issue. That may have been fixed with just setting the `writable` key to true.
in the PR we might write to apps directory that are not supposed to be written. That should be an error as it was imo.

andrey18106:
> From [#618 (comment)](https://github.com/nextcloud/app_api/issues/618#issuecomment-3124541393) and https://docs.nextcloud.com/server/latest/admin_manual/apps_management.html#using-custom-app-directories, it seems like `apps_path` is always supposed to be an array but it wasn't in the linked issue so it was a config issue. That may have been fixed with just setting the `writable` key to true. in the PR we might write to apps directory that are not supposed to be written. That should be an error as it was imo.

I agree, lets then keep it as it was, since it is supposed to work with apps_paths that is common recommended way of configuration.

PR #653, opened 2025-09-03
fix(migration): missing attributes

ArtificialOwl:
image


oleksandr-nc (approved):
lgtm. Do we need to backport this to Nextcloud 32?

PR #656, opened 2025-09-04
[stable32] fix CI after branch off

oleksandr-nc:


kyteinsky (approved):
we could also branch off this app to stable32 and update the main branch's version and server compat to 33

PR #702, opened 2025-11-14
Enh/noid/strings

julien-nc:
* Adjust a few user-facing strings
* Remove placeholders on NcCheckboxRadioSwitch
* Add mandatory names on NcModals
* Make a cancel button less emphasized


oleksandr-nc (approved):
Looks nice!

PR #713, opened 2025-11-22
fix: Use OCP\IAppConfig instead of OCP\IConfig to prevent config error

R0Wi:
When using the CLI to register a daemon

```
php occ app_api:daemon:register --set-default --harp --harp_frp_address "appapi-harp:8782" --harp_shared_key "<pass>" "harp_proxy_docker" "HaRP Proxy (Docker)" "docker-install" "http" "appapi-harp:8780" "http://my-nc-instance.com"
```

I'm getting:

```
In AppConfig.php line 873:
                                                           
  conflict between new type (mixed) and old type (string)
```

I think the problem only occurs when I already had a previous daemon installed which I uninstalled first.

image

This PR does not only resolve the problem but also removes the use of a deprecated NC function.

### DIFF
--- a/lib/Command/Daemon/RegisterDaemon.php
+++ b/lib/Command/Daemon/RegisterDaemon.php
@@ -12,7 +12,7 @@ namespace OCA\AppAPI\Command\Daemon;
 use OCA\AppAPI\AppInfo\Application;
 use OCA\AppAPI\Service\DaemonConfigService;
 
-use OCP\IConfig;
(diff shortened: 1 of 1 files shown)


oleksandr-nc:
We need a fix for this fix. `default_daemon_config` treated everywhere as `lazy` setting.

Will open a PR for this.

oleksandr-nc:
It's better to do it with two backports so as not to lose the `git` history

Issue #697, opened 2025-11-10
Setup check for the HaRP container's version


kyteinsky:
### Describe why it is important and where it will be useful

We don't have any automation in updates of the manually installed deploy daemons. It is easy to forget the updates and then face 403 issues for example when new rules are added to the deploy daemon and app_api expects them there but the daemon is too old.

### Describe your proposed solution

It would be good to have a minimum supported version set in app_api and do periodic setup checks, maybe once a week, to check if the all the deploy daemons are at least that version.
Unfortunately, the ship has sailed for Docker Socket Proxy (DSP) but HaRP does have a route for that: https://github.com/nextcloud/HaRP/blob/190adfa070a7450fe0278b4b853ba566f2e4313c/haproxy_agent.py#L517-L518
The setup check should therefore only perform this check on the HaRP daemons.

oleksandr-nc:
Closing as **complete** by the linked PR

PR #739, opened 2025-12-30
#697 Setup check for the HaRP container's version

bnw:
Hi there,
I would like to help with this ticket.
This is my first contribution to Nextcloud, so I might need some pointers :)

Some open questions I already have:

1. I'm using the built-in [version-compare](https://www.php.net/manual/en/function.version-compare.php). I'm guessing the HaRP version schema follows the PHP version string formatting closely enough for now. Do you have any concerns about this?
2. I set the MINIMUM_HARP_VERSION to the current HaRP version 0.3. Is this appropriate?
3. I'm caching the HaRP version checks for a week to keep the setup warnings page snappy. I'm using crc32 to quickly hash the HaRP settings into a cache key. I'm not familiar with the Entity system. Is there maybe a more elegant way to hash an Entity or even attach cache values to it directly? 
4. Do you think there should be more infos (maybe in the docs) than what the warning messages give?
5. I wasn't sure how to add tests for this. I also couldn't find any tests for the already existing SetupCheck (`DaemonCheck`). If you think it is necessary to add tests, could you give me some pointers on how to set this up? 
6. Do I need to update the translation files manually? Or is there some script to update them automatically? If yes, could you point me to it?

Thanks a lot and let me know if you require any changes :)

### DIFF
--- a/lib/SetupChecks/HarpVersionCheck.php
+++ b/lib/SetupChecks/HarpVersionCheck.php
@@ -78,7 +78,7 @@ class HarpVersionCheck implements ISetupCheck {
 
 		if (!empty($issues)) {
 			return SetupResult::warning(
-				implode('\n', $issues),
+				implode("<br />", $issues),
 				'https://github.com/nextcloud/HaRP/',
 			);
 		}


lib/AppInfo/Application.php
```diff
@@ -42,6 +43,7 @@ class Application extends App implements IBootstrap {
 	public const APP_ID = 'app_api';
 	public const TEST_DEPLOY_APPID = 'test-deploy';
 	public const TEST_DEPLOY_INFO_XML = 'https://raw.githubusercontent.com/nextcloud/test-deploy/main/appinfo/info.xml';
+	public const MINIMUM_HARP_VERSION = '0.3';
```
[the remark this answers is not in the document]
kyteinsky: *only if https://github.com/nextcloud/app_api/pull/739#discussion_r2660663096 change is implemented*

kyteinsky:
> I already did and it works fine locally :)
> I started with https://github.com/juliusknorr/nextcloud-docker-dev, cloned app_api into the apps-extra folder, and added a harp docker container. Took some time to figure out, but I could observe the version check coming up on the "Setups warnings" page :)

perfect!

lib/SetupChecks/HarpVersionCheck.php
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	public function getCategory(): string {
+		return 'system';
+	}
+
+	/**
+	 * @return DaemonConfig[]
+	 */
+	public function getHaRPDaemonConfigs(): array {
```
oleksandr-nc: This method is public but only used internally. Should we consider making it private?

lib/SetupChecks/HarpVersionCheck.php
```diff
@@ -0,0 +1,103 @@
...
+				}
+			} catch (\Exception $e) {
+				$this->logger->error('Failed to check HaRP version for daemon ' . $daemonConfig->getName() . ': ' . $e->getMessage(), ['exception' => $e]);
+				$issues[] = $this->l10n->t('Failed to check HaRP version for daemon "%s": %s', [$daemonConfig->getName(), $e->getMessage()]);
+			}
+		}
+
+		if (!empty($issues)) {
+			return SetupResult::warning(
+				implode('\n', $issues),
```
oleksandr-nc: Image

lib/SetupChecks/HarpVersionCheck.php
bnw: :+1:

lib/SetupChecks/HarpVersionCheck.php
```diff
@@ -0,0 +1,103 @@
...
+use OCA\AppAPI\Service\DaemonConfigService;
+use OCA\AppAPI\Service\HarpService;
+use OCP\ICacheFactory;
+use OCP\IL10N;
+use OCP\SetupCheck\ISetupCheck;
+use OCP\SetupCheck\SetupResult;
+use Psr\Log\LoggerInterface;
+
+class HarpVersionCheck implements ISetupCheck {
+	private \OCP\ICache $cache;
```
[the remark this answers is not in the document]
bnw: 👍

bnw (commented):
@oleksandr-nc Thanks a lot for your review, and sorry for the bugs. I have addressed all of them now.

While doing so, I got one more concern about security:
In case a HaRP version is too old, I'm printing the version reported by the HaRP server. This string is not escaped at all could probably inject a script tag. I'm not familiar with Nextcloud's security mechanisms and if it prevents this via content security policy headers. 
And I'm also not sure if this is really a problem, because maybe if the HaRP server (or the DNS entry to it) is compromised, all is lost anyways?
But if you think this is a problem, could you point me to the right escaping mechanisms?

oleksandr-nc (approved):
Looks good! 👍

PR #741, opened 2026-01-08
fix(daemon-name): check for forbidden character in daemon name

oleksandr-nc:
Fixes a validation flaw where daemon names containing control characters (newlines, tabs, null bytes) could bypass uniqueness checks and create undeletable entries.

Adds validation in registerDaemonConfig() and updateDaemonConfig() to reject names containing ASCII control characters (0x00-0x1F, 0x7F).

### DIFF
--- a/lib/Service/DaemonConfigService.php
+++ b/lib/Service/DaemonConfigService.php
@@ -31,7 +31,21 @@ class DaemonConfigService {
 	) {
 	}
 
+	/**
+	 * Validate that a string does not contain control characters.
+	 * Control characters (0x00-0x1F and 0x7F) can cause issues with URL routing and display.
+	 */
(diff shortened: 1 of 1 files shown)


lib/Service/DaemonConfigService.php
```diff
@@ -31,7 +31,21 @@ public function __construct(
 	) {
 	}
 
+	/**
+	 * Validate that a string does not contain control characters.
+	 * Control characters (0x00-0x1F and 0x7F) can cause issues with URL routing and display.
+	 */
+	public function containsControlCharacters(string $value): bool {
```
kyteinsky:
```suggestion
	private function containsControlCharacters(string $value): bool {
```

PR #746, opened 2026-01-13
fix(hint): explain that "--force" option exists

### LINKED ISSUE #624: Add a hint about the `--force` options when unregistering an app

Sometimes ex-app containers are unresponsive or have been already deleted using docker commands. The `occ app_api:app:unregister` command fails with and similar errors:
```
Client error: `DELETE https://exserver.myhost.tld:2375/v1.41/containers/nc_app_my_ex_app?force=true` resulted in a `404 Not Found` response:
{"message":"No such container: nc_app_my_ex_app"}


Failed to remove ExApp my_ex_app container
```

This can be easily fixed by using the `--force` option with the unregister command but it would be nice to have that hint in the above error message, better yet separate error messages can be shown when a "404 Not Found" response is received from the container deletion and in other cases.
A warning to also do a manual cleanup of the container "nc_app_<ex_app_id>" can help too.

_No response_

### Additional context

_No response_

oleksandr-nc:
Fixes #624

When the `occ app_api:app:unregister` command fails to remove a container (e.g., because it was already deleted manually), users are now shown a helpful hint suggesting the `--force` option to continue unregistering.

The hint includes the container name for easier manual verification/cleanup.

Before:
```
  Failed to remove ExApp my_ex_app container
```

After:
```
  Failed to remove ExApp my_ex_app container
  Hint: If the container "nc_app_my_ex_app" was already removed manually, you can use the --force option to continue unregistering.
  
```

### DIFF
--- a/lib/Command/ExApp/Unregister.php
+++ b/lib/Command/ExApp/Unregister.php
@@ -120,7 +120,7 @@ class Unregister extends Command {
 				if ($removeResult) {
 					if (!$silent) {
 						$output->writeln(sprintf('Failed to remove ExApp %s container', $appId));
-						$output->writeln(sprintf('Hint: If the container "%s" was already removed manually, you can use the --force option to continue unregistering.', $containerName));
+						$output->writeln(sprintf('Hint: If the container "%s" was already removed manually, you can use the --force option to fully remove it from AppAPI.', $containerName));
 					}
 					if (!$force) {
 						return 1;


lib/Command/ExApp/Unregister.php
```diff
@@ -112,12 +113,14 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
 			} else {
+				$containerName = $this->dockerActions->buildExAppContainerName($appId);
 				$removeResult = $this->dockerActions->removeContainer(
-					$this->dockerActions->buildDockerUrl($daemonConfig), $this->dockerActions->buildExAppContainerName($appId)
+					$this->dockerActions->buildDockerUrl($daemonConfig), $containerName
 				);
 				if ($removeResult) {
 					if (!$silent) {
 						$output->writeln(sprintf('Failed to remove ExApp %s container', $appId));
+						$output->writeln(sprintf('Hint: If the container "%s" was already removed manually, you can use the --force option to continue unregistering.', $containerName));
```
kyteinsky:
```suggestion
						$output->writeln(sprintf('Hint: If the container "%s" was already removed manually, you can use the --force option to fully remove it from AppAPI.', $containerName));
```

kyteinsky (approved):
🚀

PR #786, opened 2026-02-12
chore: improve haproxy password UX in daemon details modal

### LINKED ISSUE #628: Backstream entered HaProxy-Passwort

At the moment after entering a password it only displays "dummySecret123" while having the real password in the database. This makes it one the one side hard to check if the correct password is set and can also lead to confusion because you may think it overrided the entered password. 

After clicking on the eye it should show the real password. 

_No response_

### Additional context

_No response_

kyteinsky:
resolves #628

image

image


kyteinsky:
ah missed cleaning that up, yeah not used

Issue #785, opened 2026-02-12
Add deploy config info to the `occ app_api:daemon:list` command output


kyteinsky:
### Describe why it is important and where it will be useful

It will be useful to get complete details of the daemon, deploy config seems missing.
Maybe it can be added as a json in a column and behind a `--show-deploy-config` flag to the occ command to not pollute the default table view.

### Describe your proposed solution

/

### Describe alternatives you've considered, if relevant

Screenshot of the daemon details in Admin settings -> AppAPI -> Click on any registered daemon

isaaclee12:
Addressing this in [PR #791](https://github.com/nextcloud/app_api/pull/791).

I added the full DeployConfig as JSON (with `haproxy_password` masked) that displays only when using `--show-deploy-config`. 

I noticed that, since some DeployConfig values already appear in the default table, this creates redundancy. Let me know in the PR if you'd prefer an alternative approach, maybe we could show no DeployConfig data in the default table, only display it with the flag.

PR #791, opened 2026-02-19
feat(daemon): add --show-deploy-config flag to daemon:list command

isaaclee12:
Relevant Issue: https://github.com/nextcloud/app_api/issues/785

This is my first PR to Nextcloud!

NOTES:

- The force push below is due to the DCO check failing. I ran `git commit --amend -s` and `git push --force-with-lease origin feat/785/add-deploy-config-info` since there's only one commit from me.

- I have yet to set up `nextcloud-docker-dev` on my machine, so I haven't been able run tests on this yet. I'm hoping I can see the results in GitHub Actions once those workflows are kicked off.

### DIFF
--- a/lib/Command/Daemon/ListDaemons.php
+++ b/lib/Command/Daemon/ListDaemons.php
@@ -9,16 +9,15 @@ declare(strict_types=1);
 
 namespace OCA\AppAPI\Command\Daemon;
 
+use OC\Core\Command\Base;
 use OCA\AppAPI\AppInfo\Application;
 use OCA\AppAPI\Service\DaemonConfigService;
-
 use OCP\IAppConfig;
-use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Helper\Table;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
-class ListDaemons extends Command {
+class ListDaemons extends Base {
 
 	public function __construct(
 		private readonly DaemonConfigService $daemonConfigService,
@@ -28,6 +27,7 @@ class ListDaemons extends Command {
 	}
 
 	protected function configure(): void {
+		parent::configure();
 		$this->setName('app_api:daemon:list');
 		$this->setDescription('List registered daemons');
 	}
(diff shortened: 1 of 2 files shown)


isaaclee12:
Hey Oleksandr! Thanks for the thorough feedback. It's been a while since I've touched PHP, so I really appreciate the pointers.

To address your feedback, I've done the following in the latest commit:
- Made the code style changes you mentioned (e.g., inline braces, idiomatic array push syntax)
- Output all `deployConfig` fields, except for `haproxy_password` which I mask as "***"
- Skipped the table entirely when the `--show-deploy-config` flag is passed, outputting each daemon as a formatted JSON block

I managed to get `nextcloud-docker-dev` up and running, and was able to test this function. Here's the output:

Without the  `--show-deploy-config` flag:
image

With the `--show-deploy-config` flag:
```
Registered ExApp daemon configs:
[
    "*",
    "docker_sock_proxy",
    "Docker Socket Proxy",
    "docker-install",
    "http",
    "nextcloud-appapi-dsp-http:2375",
    {
        "net": "master_default",
        "nextcloud_url": "http://nextcloud.local",
        "haproxy_password": "***",
        "computeDevice": {
            "id": "cpu",
            "label": "CPU"
        },
        "harp": null
    }
]
[
    "",
    "docker_local",
    "Docker Local Socket",
    "docker-install",
    "http",
    "/var/run/docker.sock",
    {
        "net": "master_default",
        "nextcloud_url": "http://nextcloud.local",
        "haproxy_password": "***",
        "computeDevice": {
            "id": "cpu",
            "label": "CPU"
        },
        "harp": null
    }
]
[
    "",
    "manual_install",
    "Manual Install",
    "manual-install",
    "http",
    "null",
    {
        "net": "host",
        "nextcloud_url": "http://nextcloud.local",
        "haproxy_password": "***",
        "computeDevice": {
            "id": "cpu",
            "label": "CPU"
        },
        "harp": null
    }
]
```

Also, in terms of future-proofing, I was wondering what you'd think about the idea to add a list of secret-related keys to mask (see the code snippet below for an example) in the `deployConfig` JSON output, to ensure they're accounted for in the future. I'm not sure if this is something that would happen, though, so let me know if this is out of scope.
```
// IMPORTANT: If any other secret variables are added to the deploy config,
// add them to $secretKeys ensure the displayed JSON is secure by default.
$secretKeys = [
'haproxy_password',
'harp_shared_key',
'password',
'token',
'secret',
];

foreach ($secretKeys as $key) {
    if (isset($deployConfig[$key])) {
        $deployConfig[$key] = '***'; 
    }
}
```

isaaclee12:
Ok, here's the changes I've made this time:

- Changed to using a --output=json flag
- The JSON-formatted output now uses an array of each daemon's JSONs. I believe this should make this output machine-readable and reusable with other systems, and also removes the need to have a separate `writeln()` call for each daemon's JSON.
- Added input validation for the `--output` flag to explicitly require either `table` or `json` as a value, otherwise throwing an `<error>` (which matches Nextcloud's error message style).
- I tested locally to ensure that results of the following inputs for this flag work as such:
 
No `--output` flag present --> Table Output by default
`--output=table` --> Table Output
`--output=json` --> JSON Output
`--output=invalid` --> Error Message
`--output=jsn` --> Error Message
`--output=` --> Error Message

- Also, running  `occ app_api:daemon:list --help` now outputs the following text in regards to the `--output` flag:
`-o, --output=OUTPUT   Output format (json, table) [default: "table"]`

And finally, all commits are squashed and rebased on main.

Issue #803, opened 2026-02-27
ExApp upgrade fails due to a init status report rejected as unauthorized


kyteinsky:
### Describe the bug

{"timestamp": "2026-02-23T20:27:25.387258+00:00", "level": "INFO", "logger": "ccb.controller", "message": "App disabled", "filename": "[controller.py](http://controller.py/)", "function": "enabled_handler", "line": 81, "thread_name": "AnyIO worker thread", "pid": 77, "version": "5.2.0"}
{"timestamp": "2026-02-23T21:12:34.004647+00:00", "level": "INFO", "logger": "uvicorn.error", "message": "Started server process [104]", "filename": "[server.py](http://server.py/)", "function": "_serve", "line": 92, "thread_name": "MainThread", "pid": 104, "version": "5.3.0", "color_message": "Started server process [\u001b[36m%d\u001b[0m]"}
{"timestamp": "2026-02-23T21:12:34.005008+00:00", "level": "INFO", "logger": "uvicorn.error", "message": "Waiting for application startup.", "filename": "[on.py](http://on.py/)", "function": "startup", "line": 48, "thread_name": "MainThread", "pid": 104, "version": "5.3.0"}
{"timestamp": "2026-02-23T21:12:34.517840+00:00", "level": "INFO", "logger": "ccb.controller", "message": "App enable state at startup: False", "filename": "[controller.py](http://controller.py/)", "function": "lifespan", "line": 91, "thread_name": "MainThread", "pid": 104, "version": "5.3.0"}
{"timestamp": "2026-02-23T21:12:34.518918+00:00", "level": "INFO", "logger": "uvicorn.error", "message": "Application startup complete.", "filename": "[on.py](http://on.py/)", "function": "startup", "line": 62, "thread_name": "MainThread", "pid": 104, "version": "5.3.0"}
{"timestamp": "2026-02-23T21:12:34.519428+00:00", "level": "INFO", "logger": "uvicorn.error", "message": "Uvicorn running on http://127.0.0.1:23001/ (Press CTRL+C to quit)", "filename": "[server.py](http://server.py/)", "function": "_log_started_message", "line": 224, "thread_name": "MainThread", "pid": 104, "version": "5.3.0", "color_message": "Uvicorn running on \u001b[1m%s://%s:%d\u001b[0m (Press CTRL+C to quit)"}
[...]

PR #807, opened 2026-03-02
fix: ExApp upgrade 401 on set_init_status

oleksandr-nc:
Related github issue: #803

Changes proposed in this PR:

- Check `enableExAppInternal()` return value in `dispatchExAppInitInternal` and abort if it fails
- Allow disabled ExApps to access `/ex-app/status` during install/update (protected on server-side with help of `status.type`)
- Add PHPUnit tests for both fixes
- Add PHPUnit CI workflow

### DIFF
--- a/lib/Service/AppAPIService.php
+++ b/lib/Service/AppAPIService.php
@@ -309,7 +309,7 @@ class AppAPIService {
 					$this->logger->error(sprintf('Error getting path info. Error: %s', $e->getMessage()), ['exception' => $e]);
 					return false;
 				}
-				if (($this->sanitizeOcsRoute($path) !== '/apps/app_api/ex-app/state') && !$exApp->getEnabled()) {
+				if (!$exApp->getEnabled() && !$this->isExemptFromEnabledCheck($path, $exApp)) {
 					$this->logger->error(sprintf('ExApp with appId %s is disabled (%s)', $request->getHeader('EX-APP-ID'), $request->getRequestUri()));
 					return false;
 				}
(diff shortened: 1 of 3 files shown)


lib/Service/AppAPIService.php
```diff
@@ -368,6 +368,28 @@ private function sanitizeOcsRoute(string $route): string {
...
+	 */
+	private function isExemptFromEnabledCheck(string $path, ExApp $exApp): bool {
+		$sanitizedPath = $this->sanitizeOcsRoute($path);
+		if ($sanitizedPath === '/apps/app_api/ex-app/state') {
+			return true;
+		}
+		$status = $exApp->getStatus();
+		$isInitializing = in_array($status['type'] ?? '', ['install', 'update'], true);
+		if ($isInitializing) {
+			if ($sanitizedPath === '/apps/app_api/ex-app/status') {
```
kyteinsky: these two conditions may be combined

tests/php/Service/AppAPIServiceTest.php
```diff
@@ -0,0 +1,293 @@
...
+			->method('post');
+
+		$this->service->dispatchExAppInitInternal($exApp);
+	}
+
+	/**
+	 * Test that dispatchExAppInitInternal proceeds normally when enable succeeds.
+	 */
+	public function testDispatchExAppInitInternalProceedsOnEnableSuccess(): void {
+		$exApp = $this->createExApp('context_chat_backend', false);
```
kyteinsky: maybe this can be left as "test_app"

PR #839, opened 2026-04-21
fix(occ): preserve ExApp deploy options on update

### LINKED ISSUE #808: ExApp upgrade fails to carry forward the previously set environment variables
### Describe the bug

/

### Steps/Code to Reproduce

using the example of live_transcription

1. have an older version of the live_transcription app setup with env vars set in "Deploy Options" or through the occ register command
2. upgrade the app through `occ app_api:app:update live_transcription`
3. see the error logs in the docker container "ValueError: Missing environment variables: LT_HPB_URL, LT_INTERNAL_SECRET"
`docker logs -n300 nc_app_live_transcription`

### Expected Results

/

### Actual Results

/

### Setup configuration

nextcloud = 33.0.0

oleksandr-nc:
Fixes #808.

`Update.php` reads `ex_deploy_options` without an appId filter. `formatDeployOptions()` then flattens by type, so whichever app's row comes last in iteration order clobbers the rest and the target app's saved env vars get dropped whenever another ExApp has rows in the table.

Tests: unit coverage in `ExAppDeployOptionsServiceTest`, a Docker integration job in `tests-deploy.yml`, and a K8s integration step in `test_occ_commands_k8s.py` that runs across all four expose-type workflows.

### DIFF
--- a/.github/workflows/tests-deploy-k8s.yml
+++ b/.github/workflows/tests-deploy-k8s.yml
@@ -216,11 +216,244 @@ jobs:
           path: data/nextcloud.log
           if-no-files-found: warn
 
+  k8s-update-preserves-deploy-options:
+    runs-on: ubuntu-22.04
+    name: Update preserves deploy options (K8s)
+    # Regression test for https://github.com/nextcloud/app_api/issues/808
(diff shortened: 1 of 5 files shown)


kyteinsky (approved):
🚀

PR #866, opened 2026-05-12
feat(talkbot): move TalkBot bookkeeping to dedicated ex_apps_talk_bots table

oleksandr-nc:
## Summary

- **New dedicated table `oc_ex_apps_talk_bots`** (`id`, `appid`, `route`, `secret`, `created_time`) replaces the two-row-per-bot K/V pattern in `appconfig_ex`. Unique index on `(appid, route)` is the natural key for register/unregister/proxy lookups; `appid` index supports per-ExApp fan-out.
- **`TalkBotsService` rewritten on top of `TalkBotMapper` / `TalkBot` entity.** Bot secrets are encrypted at rest via `ICrypto`. Re-register is a no-op on AppAPI's side - Talk owns the human-readable metadata via `BotInstallEvent` and reconciles its own row on every register.
- **One-shot data migration backfills existing bots from `appconfig_ex`.** Handles both plaintext (the legacy default, since the old service never passed `$sensitive=true`) and `sensitive=1` encrypted secret rows. Re-encrypts on insert into the new table. Each bot runs in its own transaction with per-bot rollback, and the migration is safe to re-run (skips bots already present in the new table).
- **OCS contract for ExApps is unchanged.** `POST/DELETE /ex-app/talk_bot` keep the same params and the same `{id, secret}` response shape. `bot_id = sha1(\$appid . '_' . \$route)` is preserved, so existing community bots (including nc_py_api callers) continue to work post-upgrade without re-registration. The proxy URL and HMAC verification path are byte-identical.
- **Fan-out cleanup on ExApp uninstall.** `ExAppService::unregisterExApp` calls `TalkBotsService::unregisterExAppTalkBots`, which dispatches `BotUninstallEvent` per bot so Talk's `talk_bots_server` stays aligned with our table.
- **Compensating uninstall on insert failure.** If `mapper->insert` fails after Talk has already persisted its row from the install event, the service fires `BotUninstallEvent` to avoid leaving Talk with an orphan pointing at a secret AppAPI never stored.
- **PHPUnit coverage** - controller register/unregister/re-register/corrupted-secret/fan-out, plus migration coverage (plaintext secret, sensitive-encrypted secret, orphan route row, malformed key suffix, multi-bot multi-app, idempotent re-run, re-run after pre-existing).

## Test plan

- [x] Full PHPUnit suite passes (68 tests, 216 assertions)
- [x] `psalm` clean (stub added for spreed `Bot{Install,Uninstall}Event`)
- [x] End-to-end: register bot via nc_py_api -> simulate pre-upgrade legacy `appconfig_ex` state with the same secret bytes -> run migration -> real HMAC-SHA256-signed POST through `proxyTalkMessage` is accepted (HMAC verification succeeds against the migrated secret)
- [x] Fan-out via `occ app_api:app:unregister`: both `oc_ex_apps_talk_bots` and `oc_talk_bots_server` rows for the ExApp are cleaned
- [x] Reviewer: smoke-test with a real Talk bot ExApp (e.g. summary_bot) on a vanilla instance - install on main, upgrade to this branch, confirm the bot still responds to messages without re-registering

## Summary by CodeRabbit

* **Infrastructure**
  * Talk bot configurations now use dedicated database storage for improved reliability and a new migration consolidates legacy configs into the new table.

* **Bug Fixes**
  * Improved secret handling: encrypted storage, decryption safeguards, corruption detection, and safer rollback on failures.

* **Behavior**
  * Bot registration/unregistration now handles duplicates, per-bot errors, and compensating cleanup reliably.

* **Tests**
  * Expanded test coverage for registration, cleanup, migration idempotency, and secret handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nextcloud/app_api/pull/866)

### DIFF
--- a/lib/Db/TalkBot.php
+++ b/lib/Db/TalkBot.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\AppAPI\Db;
+
+use JsonSerializable;
+use OCP\AppFramework\Db\Entity;
+
+/**
+ * Class TalkBot
+ *
+ * @package OCA\AppAPI\Db
+ *
+ * @method string getAppid()
+ * @method string getRoute()
+ * @method string getSecret()
+ * @method int getCreatedTime()
+ * @method void setAppid(string $appid)
+ * @method void setRoute(string $route)
+ * @method void setSecret(string $secret)
+ * @method void setCreatedTime(int $createdTime)
+ */
+class TalkBot extends Entity implements JsonSerializable {
+	protected $appid;
+	protected $route;
+	protected $secret;
+	protected $createdTime;
+
+	public function __construct() {
+		$this->addType('appid', 'string');
+		$this->addType('route', 'string');
+		$this->addType('secret', 'string');
+		$this->addType('createdTime', 'integer');
+	}
+
+	public function jsonSerialize(): array {
+		return [
+			'id' => $this->getId(),
+			'appid' => $this->getAppid(),
+			'route' => $this->getRoute(),
+			'created_time' => $this->getCreatedTime(),
+		];
+	}
+}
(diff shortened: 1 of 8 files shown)


edward-ly (approved):
Nice! Works with `talk_bot_ai`

PR #874, opened 2026-05-15
fix: proxy route leading slash

oleksandr-nc:
Restores route matching for ExApps that declare routes in the canonical regex form documented in HaRP (`<url>^/path$</url>`, e.g. [app-skeleton-python](https://github.com/nextcloud/app-skeleton-python)).

`ExAppProxyController::passesExAppProxyRoutesChecks` builds a `^`-anchored regex from `$route['url']` and matches it against `$other`.
Symfony strips the leading slash, so `$other` is bare (e.g. `public/page`). But route URLs are typically written against the full path.

To avoid breaking ExApps whose info.xml still uses the non-canonical bare form (currently context_chat_backend's `<url>downloadLogs</url>`), the second commit adds a one-iteration fallback: if the canonical subject didn't match, retry against the bare subject and emit a debug log identifying the ExApp and route so maintainers know to migrate. Marked _TODO(deprecation)_ for removal once known legacy apps have shipped the canonical form.

### DIFF
--- a/lib/Controller/ExAppProxyController.php
+++ b/lib/Controller/ExAppProxyController.php
@@ -318,10 +318,19 @@ class ExAppProxyController extends Controller {
 
 	private function passesExAppProxyRoutesChecks(ExApp $exApp, string $exAppRoute): array {
 		// Route URL is a regex matched against the request path including its leading slash, mirroring HaRP's target_path semantics.
-		$exAppRoute = '/' . $exAppRoute;
+		$canonicalSubject = '/' . $exAppRoute;
 		foreach ($exApp->getRoutes() as $route) {
 			$pattern = '~^(?:' . str_replace('~', '\\~', $route['url']) . ')~i';
-			if (preg_match($pattern, $exAppRoute) === 1
+			$matched = preg_match($pattern, $canonicalSubject) === 1;
+			if (!$matched && preg_match($pattern, $exAppRoute) === 1) {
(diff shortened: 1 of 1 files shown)


lib/Controller/ExAppProxyController.php
```diff
@@ -317,9 +317,20 @@ private function buildMultipartFormData(array $bodyParams, array $files): array
...
 
 	private function passesExAppProxyRoutesChecks(ExApp $exApp, string $exAppRoute): array {
+		// Route URL is a regex matched against the request path including its leading slash, mirroring HaRP's target_path semantics.
+		$canonicalSubject = '/' . $exAppRoute;
 		foreach ($exApp->getRoutes() as $route) {
 			$pattern = '~^(?:' . str_replace('~', '\\~', $route['url']) . ')~i';
-			if (preg_match($pattern, $exAppRoute) === 1
+			$matched = preg_match($pattern, $canonicalSubject) === 1;
+			if (!$matched && preg_match($pattern, $exAppRoute) === 1) {
+				// TODO(deprecation): remove this bare-path fallback once known ExApps have migrated their info.xml route URLs to the canonical `^/path$` form. Tracked by the AppAPI / context_chat_backend coordination effort.
```
kyteinsky, line 326: not sure if any other ex-app is using this still but would be good to mark a breaking change in the changelog if this is done sometime.

PR #885, opened 2026-05-16
feat: validate ExApp routes at registration and reject malformed data

oleksandr-nc:
Follow-up to #882, which added defensive read-side parsing for nullable / malformed `bruteforce_protection` and `headers_to_exclude` columns. 
That PR which added defensive read-side parsing for nullable / malformed `bruteforce_protection` and `headers_to_exclude` columns. That PR kept legacy rows working but left the write side silently coercing any unrecognised value to `'[]'` - so a developer with a typo in their `info.xml` got an apparently-installed ExApp whose routes didn't behave as written, with no error to point them at the mistake.

This PR moves to fail-fast: every route definition from `--info-xml`, `--json-info` runs through a new `ExAppRouteHelper::normalizeAndValidate()` step inside `ExAppService::getAppInfo()`.

### DIFF
--- a/lib/Service/ExAppRouteHelper.php
+++ b/lib/Service/ExAppRouteHelper.php
@@ -1,0 +1,158 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\AppAPI\Service;
+
+use InvalidArgumentException;
+
+/**
+ * Normalize and validate ExApp routes from info.xml / --json-info before they are persisted.
+ *
+ * Two input shapes feed this helper:
+ *   - JSON (`--json-info`): values are already typed (access_level: int, bruteforce_protection: int[], headers_to_exclude: string[]).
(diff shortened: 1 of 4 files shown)


lib/Service/ExAppRouteHelper.php
```diff
@@ -0,0 +1,158 @@
...
+			return [];
+		}
+		$out = [];
+		foreach ($list as $index => $value) {
+			if (!is_int($value)) {
+				throw new InvalidArgumentException(sprintf("%s: '%s' must contain only integers (e.g. HTTP status codes), entry at index %d is %s", $ident, $field, $index, self::describe($value)));
+			}
+			$out[] = $value;
+		}
+		return $out;
```
kyteinsky, line 102: maybe this can be checked with `array_all()` and then the `$list` can be just returned instead of re-creating it
same for `normalizeStringList()`

kyteinsky (approved):

lib/Service/ExAppRouteHelper.php
oleksandr-nc: Screenshot From 2026-05-19 10-21-48

**PHP8.4** 😭

kyteinsky: ah damn, missed that part :sweat_smile:

PR #887, opened 2026-05-18
feat: add default_action field to files-actions-menu registration

### LINKED ISSUE #748: Support default file action registration

I'm working on integrating IronCalc for editing XLSX files with Nextcloud and while it's cool that I can add things to the files actions menu, the default action is still 'download.'

An extra field added to https://docs.nextcloud.com/server/stable/developer_manual/exapp_development/tech_details/api/fileactionsmenu.html#register like `"default": true | false`.

Not aware of any viable alternatives, besides creating an additional app just to do this in PHP directly. Or is there a way to inject a piece of PHP in an ExApp?

### Additional context

PS. sometimes (seemingly randomly) my file actions aren't showing up. Is this a known bug? There's no errors I can see. It happens most with Chrome inspector open with disable cache.
PPS. I think `"mime": "file"` doesn't work; when I use that my actions never show up.

jvdp:
fixes #748


oleksandr-nc:
@jvdp , could you rebase the PR when you have time?

jvdp:
@oleksandr-nc rebase done and also fixed up a bit. The second commit is to fix a bug found on CI.

lib/Service/UI/FilesActionsMenuService.php
```diff
@@ -44,39 +44,37 @@ public function __construct(
...
+			$entity->setName($name);
+			$entity->setDisplayName($displayName);
+			$entity->setActionHandler(ltrim($actionHandler, '/'));
+			$entity->setIcon(ltrim($icon, '/'));
+			$entity->setMime($mime);
+			$entity->setPermissions((string)$permissions);
+			$entity->setOrder($order);
+			$entity->setVersion($version);
+			$entity->setDefaultAction($defaultAction);
 			$this->resetCacheEnabled();
```
oleksandr-nc, line 71: Could we move `resetCacheEnabled()` to AFTER the DB write ? Will it be more correct?

The load-then-mutate refactor is correct and fixes a real bug (null clearing now persists - nice catch), but the cache invalidation order changed in a way that introduces a sticky-stale race

The cache here is a single distributed key `/ex_ui_files_actions` holding the full enabled-list across **all** ExApps. It's read on every Files page load (`LoadFilesPluginListener::handle` --> `getRegisteredFileActions`) and lazily repopulated on miss from `findAllEnabled()`

**Before this PR (evict-after-write):**
```
T1: INSERT/UPDATE in DB
T2: cache.remove()
```
A reader landing between T1 and T2 reads cached pre-write data once, but T2 evicts immediately so the next reader misses and refetches truth

**This PR (evict-before-write):**
```
T1: cache.remove()
T2: INSERT/UPDATE in DB
```
A reader landing between T1 and T2 *misses* the cache (T1 cleared it), runs `findAllEnabled()` against the still-pre-write DB, and **re-populates the cache with stale data**. That stale entry now persists until the next register/unregister of *any* ExApp's action. Sticky stale

Real-world frequency is **low** (admin-only registrations are rare), but the fix is free and also avoids evicting the cache when the DB write throws

Suggested change:

```php
try {
    $entity = $existing ?? new FilesActionsMenu();
    $entity->setAppid($appId);
    $entity->setName($name);
    $entity->setDisplayName($displayName);
    $entity->setActionHandler(ltrim($actionHandler, '/'));
    $entity->setIcon(ltrim($icon, '/'));
    $entity->setMime($mime);
    $entity->setPermissions($permissions);
    $entity->setOrder($order);
    $entity->setVersion($version);
    $entity->setDefaultAction($defaultAction);
    $result = $existing === null ? $this->mapper->insert($entity) : $this->mapper->update($entity);
    $this->resetCacheEnabled();
    return $result;
} catch (Exception $e) {
    $this->logger->error(
        sprintf('Failed to register ExApp %s FileActionMenu %s. Error: %s', $appId, $name, $e->getMessage()),
        ['exception' => $e]
    );
    return null;
}
```

jvdp: Well spotted!

oleksandr-nc:
@jvdp the PR looks really good! ☀️

Could you squash all commits into one? After that we can merge this

oleksandr-nc (approved):

oleksandr-nc:
@jvdp , thank you for taking effort and working on this - good job.

jvdp:
@oleksandr-nc you're welcome and thanks for merging!

PR #891, opened 2026-06-05
fix(fetcher): catch GenericFileException when reading the appstore cache file

oleksandr-nc:
Mirrors the core fix in nextcloud/server#60286: on `GenericFileException` (NC 35 now throws it when the appstore cache file can't be read) we drop the unreadable cache file and let it be refetched, returning `[]` cleanly if even deleting it fails.

Tested on NC 35 by making the cached `appapi_apps.json` unreadable: before, `/apps/list` returns 500; after, it recovers with the full list and the cache file is recreated.

### DIFF
--- a/lib/Fetcher/AppAPIFetcher.php
+++ b/lib/Fetcher/AppAPIFetcher.php
@@ -15,6 +15,7 @@ use OC\Files\AppData\Factory;
 use OCA\AppAPI\AppInfo\Application;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\Files\GenericFileException;
 use OCP\Files\IAppData;
 use OCP\Files\NotFoundException;
 use OCP\Files\NotPermittedException;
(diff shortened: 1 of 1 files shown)


oleksandr-nc:
CI failures are not related

lib/Fetcher/AppAPIFetcher.php
```diff
@@ -148,6 +149,16 @@ public function get(bool $allowUnstable = false): array {
 		} catch (NotFoundException $e) {
```
davesnith4538-dot, line 149: Meaning

davesnith4538-dot: Today is 1st day things kinda vome to understanding and downloading lots around light and matta and its up and down im not sure what's what yet

lib/Fetcher/AppAPIFetcher.php
```diff
@@ -15,6 +15,7 @@
 use OCA\AppAPI\AppInfo\Application;
```
davesnith4538-dot, line 15: Yes

lib/Fetcher/AppAPIFetcher.php
oleksandr-nc: hello. can you explain a little bit the question?

davesnith4538-dot: Hi im just trying get my head arounf this hole thing  is ok to some one to help insight me on in to what im thinking this is is right?

oleksandr-nc: this PR is the clone of the server PR: https://github.com/nextcloud/server/pull/60286 to be in sync with the changes in the server

davesnith4538-dot: Ok so what shall one do ?

davesnith4538-dot:
So I shall do ??? Reports? Daily ?

Issue #888, opened 2026-05-22
Allow ExApps to create setup checks


lukasdotcom:
### Describe why it is important and where it will be useful

Sometimes an exapp has some sort of issue that should be displayed to the administrator just like with setup checks. For example could be used for live transcription to show the admin that they configured it incorrectly.

### Describe your proposed solution

Allow an exapp to register a setup check (https://docs.nextcloud.com/server/stable/developer_manual/digging_deeper/setup_checks.html) that would be shown on the admin panel.

### Describe alternatives you've considered, if relevant

Show issues with ex apps on the app api settings page instead.

oleksandr-nc:
Closing as **done**

PR #908, opened 2026-06-27
feat(setup-checks): let ExApps surface setup checks in the admin overview

oleksandr-nc:
Implements #888.

ExApps can now surface admin-facing setup checks in Settings > Administration > Overview, next to the
core ones (e.g. an ExApp warning that it isn't configured correctly). An ExApp opts in with
`POST /api/v1/setup_check` (and `DELETE` to opt out); app_api registers a single aggregate "External
Apps" `ISetupCheck`, and the results come from each opted-in ExApp's `/setup_checks` endpoint. A down
or slow ExApp shows "not responding".

The SetupCheck page itself does no HTTP. The fan-out to ExApps runs entirely in the background (a
10-min TimedJob plus a one-off job that an admin's page visit enqueues, stale-while-revalidate) and
stores the results in `IAppConfig`; the page just reads those, filtered to the currently opted-in and
enabled ExApps. So a slow, down, or errored ExApp can never slow down the admin page.

Notes:
- No new table or migration: the opt-in marker and the computed results both live in `IAppConfig`.
- ExApp text is escaped and link URLs validated before rendering; response size, string length, and entry count are bounded.
- Deploy/init errors stay out of scope (already shown on the ExApp management page).

Draft: the `nc_py_api` helper (a `@setup_check` decorator) and docs are a follow-up; this is the
server/app_api side. NC35-only, not backported.

### Screenshots

Two ExApps and each one prints `Error` + `Warning`

Screenshot From 2026-06-30 14-50-37

One ExApps prints `Error` + `Warning`:

Screenshot From 2026-06-30 15-12-04

### DIFF
--- a/lib/Service/ExAppSetupCheckRefreshService.php
+++ b/lib/Service/ExAppSetupCheckRefreshService.php
@@ -1,0 +1,204 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\AppAPI\Service;
+
+use OCA\AppAPI\Db\ExApp;
+use OCP\Http\Client\IResponse;
+use OCP\IL10N;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Polls each declaring ExApp's `/setup_checks` endpoint and stores the computed issues, so the
+ * admin "Security & setup warnings" page (see {@see \OCA\AppAPI\SetupChecks\ExAppsSetupCheck}) never
+ * does any HTTP itself. Runs only in the background (a TimedJob + an on-demand QueuedJob fired when an
+ * admin opens the page), so a slow/down/errored ExApp can never slow the page - it only costs time
+ * here. It catches all its own errors and never throws.
+ *
+ * Text produced here is in the server-default language (there is no viewer in a background job);
+ * ExApp-provided text is monolingual either way. All ExApp-provided values are length-capped and the
+ * link URL is validated; HTML-escaping happens at render time in the SetupCheck.
+ */
+class ExAppSetupCheckRefreshService {
+	private const PER_APP_TIMEOUT_SECONDS = 5;
+	private const MAX_RESPONSE_BYTES = 262144; // 256 KiB
+	private const MAX_RESPONSE_ENTRIES = 1000;
+	private const MAX_TEXT_LENGTH = 4096;
+
+	public function __construct(
+		private readonly IL10N $l10n,
+		private readonly LoggerInterface $logger,
+		private readonly ExAppSetupCheckService $setupCheckService,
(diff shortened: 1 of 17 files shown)


openapi.json
```diff
@@ -5267,6 +5267,139 @@
...
+                    {
+                        "name": "OCS-APIRequest",
+                        "in": "header",
+                        "description": "Required to be true for the API request to pass",
+                        "required": true,
+                        "schema": {
+                            "type": "boolean",
+                            "default": true
+                        }
+                    }
```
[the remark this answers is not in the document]
oleksandr-nc: no AppAPI ex-app controller documents it (it's the AppAPIAuth mechanism); doing it only here would be inconsistent

lib/Service/ExAppSetupCheckService.php
```diff
@@ -0,0 +1,103 @@
...
+		$apps = (is_array($decoded) && isset($decoded['apps']) && is_array($decoded['apps'])) ? $decoded['apps'] : [];
+		$updatedAt = (is_array($decoded) && isset($decoded['updatedAt'])) ? (int)$decoded['updatedAt'] : 0;
+		return ['apps' => $apps, 'updatedAt' => $updatedAt];
+	}
+
+	/**
+	 * @param array<string, list<array<string, mixed>>> $apps appId => list of issues
+	 */
+	public function storeState(array $apps): void {
+		$this->appConfig->setValueString(Application::APP_ID, self::STATE_KEY, json_encode(['apps' => $apps, 'updatedAt' => time()]));
```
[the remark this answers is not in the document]
oleksandr-nc: this one should be fixed

lib/Service/ExAppSetupCheckRefreshService.php
```diff
@@ -0,0 +1,204 @@
...
+		}
+		return $issues;
+	}
+
+	/**
+	 * @return array{severity: string, appName: string, text: string, linkUrl: string, linkLabel: string}
+	 */
+	private function notRespondingIssue(ExApp $exApp): array {
+		return [
+			'severity' => 'warning',
```
lukasdotcom, line 170: IMO not responding feels like an error

lukasdotcom (approved):
I didn't really find anything wrong, but didn't look that closely. The general implementation does look like what I'd want.

lib/Service/ExAppSetupCheckRefreshService.php
oleksandr-nc: Logic why this is a warning:

1. Blast radius: one opted-in ExApp being unreachable degrades only that app, not the platform, so it matches `HarpVersionCheck` (which already surfaces an unreachable daemon as a warning)
2. Usually transient: it fires for an enabled, non-initializing app that just missed the `5s` probe (restart, update tail, brief overload).
3. The real up/down status already lives on the AppAPI management page; this line is a secondary "couldn't fetch its self-reported checks" hint

lukasdotcom: Then you can leave it as is.

PR #953, opened 2026-07-29
fix(notifications): do not translate empty ExApp strings, which throws on Nextcloud 35

oleksandr-nc:
Nextcloud master throws `RuntimeException: The translated text is empty` since the L10N strict types refactoring, so a notification created without a message made `ExAppNotifier::prepare()` call `t('')` and turned the whole notifications endpoint into a 500 - which is what has been failing the NC_Py_API jobs on every PR.

Empty rich subject/message are now skipped instead of translated.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI


lukasdotcom (approved):
🐘

AndyScherzinger (approved):

PR #981, opened 2026-08-20
fix: make db migrations idempotent

kyteinsky:
## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/lib/Migration/Version032002Date20250527174907.php
+++ b/lib/Migration/Version032002Date20250527174907.php
@@ -18,6 +18,7 @@ use OCP\Migration\Attributes\ColumnType;
 use OCP\Migration\IOutput;
 use OCP\Migration\SimpleMigrationStep;
 use OCP\Security\ICrypto;
+use Throwable;
 
 #[AddColumn('preferences_ex', 'sensitive', ColumnType::SMALLINT, 'support sensitive setting')]
 class Version032002Date20250527174907 extends SimpleMigrationStep {
@@ -71,6 +72,11 @@ class Version032002Date20250527174907 extends SimpleMigrationStep {
 		while ($row = $req->fetch()) {
 			$configValue = $row['configvalue'];
 			if (!empty($configValue)) {
+				// Readers decrypt exactly once, so re-encrypting an already-encrypted value would
+				// nest a second envelope and make the stored secret unrecoverable on the next read.
+				if ($this->isEncrypted((string)$configValue)) {
+					continue;
+				}
 				try {
 					$encryptedValue = $this->crypto->encrypt($configValue);
 					$qbUpdate = $this->connection->getQueryBuilder();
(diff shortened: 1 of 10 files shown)


kyteinsky:
- https://github.com/nextcloud/app_api/blob/main/lib/Migration/Version5000Date20241120135411.php
- https://github.com/nextcloud/app_api/blob/main/lib/Migration/Version032002Date20250527174907.php

these two encrypt values in the migrations, maybe they can be made idempotent too by trying to decrypt the values to check if the value in the DB is already encrypted, but not sure if that should be done or not, or if there is a better way.

oleksandr-nc:
I tested this live on NC35 + PostgreSQL by wiping `oc_migrations` for app_api and replaying all 25 migrations. The guards do what they should: `2206` and a partially applied `2000` (index already dropped) both go from a hard failure to a clean re-run, the replay now completes 25/25, and it ends on a schema identical to the original with no row loss. Fresh installs are unaffected on both the `migrate()` and the schema-only `Installer` path.

One thing needs to land with it, though. `Version5000Date20241120135411` sorts after `2206`, so today the replay dies at `2206` and never reaches it. Once the replay completes, that step re-encrypts every daemon's `haproxy_password` a second time:

```
main:     exit=1 at 2206,  haproxy_password len=228 (untouched)
this PR:  exit=0, 25/25,   haproxy_password len=644  (double encrypted)
```

All the consumers (`HarpController::validateHarpSharedKey`, `HarpService`, `DockerActions`, `KubernetesActions`) decrypt exactly once, so a single `decrypt()` returns ciphertext and every daemon's HaRP key silently stops working. `Version032002Date20250527174907` has the same defect on `appconfig_ex` sensitive values (15 -> 196 -> 580 bytes over two runs).

I pushed 3f85de3 on top with a skip-if-already-encrypted guard for both. Verified: a plaintext password is still encrypted on the first run and decrypts back correctly, a second run is a no-op, and the full replay now finishes 25/25 with len=228. psalm and php-cs-fixer are clean.

@kyteinsky could you have a look? If it looks right to you, feel free to squash it into your commit and force push.

Two more things I noticed while testing, not covered by my commit:

* `2800Date20240711080316`, `3000Date20240807085759`, `3100Date20240822080316` and `3200Date20240905080316` still call `getTable()` unguarded, and `032000Date20250109162434` calls `getColumn()` unguarded, so they fail exactly the same way. Worth folding in, or narrowing the PR title.
* All four files here are byte-identical on stable33 and stable34, and main is unreleased NC35, so the people who can actually hit this are on the stable branches. Backports would be good.

kyteinsky:
thanks, the fix looks good!
pushed the missing guards.

oleksandr-nc (approved):
Nice!

PR #1046, opened 2026-09-21
fix(k8s): apply daemon registry mappings

### LINKED ISSUE #1039: Kubernetes deploy daemon ignores registry mappings (app_api:daemon:registry:add)
## Problem

A deploy daemon can carry registry mappings (`occ app_api:daemon:registry:add <daemon> --registry-from ghcr.io --registry-to registry.example.com`), so ExApp images are taken from a mirror or a private registry instead of the registry named in the ExApp's `info.xml`. This is what air-gapped and registry-restricted environments rely on.

The mapping is only applied by the Docker deploy backend. With a Kubernetes deploy daemon the command succeeds, `registry:list` shows the mapping, and nothing else happens: the ExApp Deployment is still created with the original image reference, so the cluster tries to pull from the upstream registry.

## Steps to reproduce

1. Register a Kubernetes deploy daemon (`kubernetes-install`).
2. `occ app_api:daemon:registry:add <daemon> --registry-from ghcr.io --registry-to registry.example.com`
3. `occ app_api:daemon:registry:list <daemon>` lists `ghcr.io -> registry.example.com`.
4. Deploy any ExApp whose image lives on `ghcr.io`, e.g. run Test Deploy.
5. `kubectl -n <exapps-namespace> get deploy -o wide` shows the image as `ghcr.io/...`, not `registry.example.com/...`.

## Expected

The Deployment is created with `registry.example.com/<image>:<tag>`, the same way the Docker backend rewrites the image before pulling.

## Cause

`DockerActions` resolves the mapping in `buildBaseImageName()`, `buildExtendedImageName()` and `shouldPullImage()`. `KubernetesActions::buildImageName()` concatenates `image_src/image_name:image_tag` from `info.xml` and never looks at `deploy_config['registries']`. The `registry:*` commands do not check the daemon type, so they accept a Kubernetes daemon without any hint that the mapping is unused there.

Affects `main`, `stable35` and `stable34` (every branch that has the Kubernetes backend).

## Proposal

- Apply the daemon's registry mapping in `KubernetesActions` when the create payload for HaRP is built, ideally through one shared helper so both backends resolve image names identically.
- Decide what the special `local` target means on Kubernetes. AppAPI does not pull there, the kubelet does, and ExApp pods use `imagePullPolicy: IfNotPresent`, so keeping the image name unchanged (a preloaded image is then used) looks like the natural equivalent.
- Cover it with a unit test for the Kubernetes image name, and mention Kubernetes in the registry mapping documentation.

## Related, to decide while at it

A mirror that needs authentication also requires pull credentials in the cluster. HaRP does not set `imagePullSecrets` on ExApp pods today, so this currently has to be solved outside (for example `imagePullSecrets` on the namespace's default ServiceAccount). Worth documenting, or supporting as a HaRP option.

## Workarounds until this is supported

- Configure a registry mirror for the upstream registry in the container runtime of the cluster nodes (containerd `hosts.toml`). Transparent for all ExApps.
- Rewrite the image registry at admission time for the ExApps namespace, e.g. with a Kyverno mutate policy ([sample](https://kyverno.io/policies/other/replace-image-registry/replace-image-registry/)).

oleksandr-nc:
Fixes #1039

Registry mappings of a deploy daemon (`occ app_api:daemon:registry:add`, or the "Docker registries" dialog) were only applied by the Docker backend. On a Kubernetes daemon the mapping was stored and listed, but `KubernetesActions` built the image name straight from `info.xml`, so the cluster kept pulling from the upstream registry. That breaks air-gapped and mirror-only setups.

## Changes

- `DaemonConfigService::resolveImageRegistry()` resolves the registry of an ExApp image from the daemon's mappings. Both Docker image name builders and `KubernetesActions::buildImageName()` use it, so the two backends resolve names the same way. On Kubernetes the daemon config is passed down to where the HaRP `/exapp/create` payload is built, for single and multi-role deployments.
- Unusable stored entries are ignored by the resolver and by `DockerActions::shouldPullImage()`. An entry is unusable when it is not an array, when `from` or `to` is missing or not a string, or when the target is empty once trailing slashes are dropped. Before, such an entry raised a `TypeError` in the middle of a deploy. On Kubernetes that would happen after the previous Deployment had already been removed.
- `addDockerRegistry()` and `removeDockerRegistry()` keep `registries` a list. Removing an entry that was not the last one left a JSON object with gaps (`{"1": {...}}`), which breaks the add form of the registries dialog. Data that already has that shape is repaired on the next add or remove.
- `addDockerRegistry()` rejects unusable mappings (same cases as above) and stores only `from` and `to`.
- `removeDockerRegistry()` no longer fails with a `TypeError` when an unusable entry is stored next to the mapping that is removed.
- `DockerActions::shouldPullImage()` follows the effective mapping. If a source has both a `local` and a mirror entry (only possible with hand-edited data, `addDockerRegistry()` rejects a duplicate source), the image name resolved to the mirror while the pull was skipped. Now the mirror image is pulled.

## Behaviour

- The `local` target has no effect on Kubernetes. AppAPI never pulls there. HaRP creates the pods with `imagePullPolicy: IfNotPresent`, so the image name is kept and the kubelet pulls it if the node does not have it. Docker is unchanged: the name is kept and the pull is skipped.
- Mappings that are already stored on a Kubernetes daemon were ignored until now. They take effect on the next ExApp install or update. The mirror has to be reachable from the cluster nodes.
- HaRP does not set `imagePullSecrets`, so a mirror that needs authentication needs pull credentials configured in the cluster.
- The GPU image variants (`<tag>-cuda`, `<tag>-rocm`) are still not used on Kubernetes.


lib/DeployActions/DockerActions.php
kyteinsky, line 460: not sure but seems like the added lines 450-452 do the same thing which the rest of the function is doing

kyteinsky (commented):
> * The `local` target has no effect on Kubernetes. AppAPI never pulls there. HaRP creates the pods with `imagePullPolicy: IfNotPresent`, so the image name is kept and the kubelet pulls it if the node does not have it. Docker is unchanged: the name is kept and the pull is skipped.

maybe we can keep the same behaviour as docker here, set to no-pull when this registry mapping is present.

